PR #94, opened 2025-11-21
[tests] add test and problem report uploading to the server test workflow

pq:
Adds test and problem report uploading to the `dart-analysis-server-tests` workflow:

image

## Sample Reports

**Tests**

image

image

**Problems**

image

---

<details>

- See our [contributor guide]([https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md](https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md) for general expectations for PRs.
- Dart contributions to our repos should follow the [Dart style guide](https://dart.dev/guides/language/effective-dart) and use `dart format`.
- Java and Kotlin contributions should strive to follow Java and Kotlin best practices ([discussion](https://github.com/flutter/flutter-intellij/issues/8098)).
</details>

### DIFF
--- a/.github/workflows/presubmit.yaml
+++ b/.github/workflows/presubmit.yaml
@@ -126,10 +126,3 @@ jobs:
           ./gradlew verifyPluginSignature
           ./gradlew verifyPluginStructure
         working-directory: third_party
-      # Collect Plugin Verifier Results
-      - name: Collect Plugin Verifier Results
-        if: ${{ always() }}
-        uses: actions/upload-artifact@v4
-        with:
-          name: pluginVerifier-results
-          path: build/reports/pluginVerifier


pq:
Hilariously, this last run has no failures, but you can check out the problems report in the uploaded zip here:

https://github.com/flutter/dart-intellij-third-party/actions/runs/19559937926/artifacts/4636302178

.github/workflows/presubmit.yaml
```diff
@@ -93,7 +93,14 @@ jobs:
         run: |
           ./gradlew :test --tests "com.jetbrains.dart.analysisServer.*"
         working-directory: third_party
-
+      - name: Upload build artifacts
+        uses: actions/upload-artifact@v4
```
jwren: If this works, this is great!  What does the CLI output look like, i.e. is it discoverable when it is written to these locations?  Also, why not add this to to the Jobs?

pq: I've been watching runs and it seems to work as intended! 🎉 

(For example, this recent run: [job/56077860294](https://github.com/flutter/dart-intellij-third-party/actions/runs/19580730530/job/56077860294) generated artifacts [here](https://github.com/flutter/dart-intellij-third-party/actions/runs/19580730530/artifacts/4644070638).)

> What does the CLI output look like, i.e. is it discoverable when it is written to these locations?

If you squint at the picture above you can see a link is printed in the results. Here it is again:

image

> Also, why not add this to to the Jobs?

You mean add a separate job? What would be the benefit?

(FWIW I did try actually try that initially but hit some snags and couldn't get the `upload-artifact` action to find the artifacts. I *think* I understand why after sleeping on it. If there's a good reason to get that working I can take another crack. Just let me know!)

PR #166, opened 2025-12-19
Lsp

helin24:
I tried commenting out places where the standard analysis server starts and instead start a server for LSP4IJ, and basic things seem to work. I haven't compared coverage or anything.

### DIFF
--- a/third_party/src/main/resources/META-INF/plugin.xml
+++ b/third_party/src/main/resources/META-INF/plugin.xml
@@ -85,7 +85,7 @@
                                implementationClass="com.jetbrains.lang.dart.ide.info.DartParameterInfoHandler"/>
 
     <langCodeStyleSettingsProvider implementation="com.jetbrains.lang.dart.ide.formatter.settings.DartLanguageCodeStyleSettingsProvider"/>
-    <lang.importOptimizer language="Dart" implementationClass="com.jetbrains.lang.dart.ide.imports.DartImportOptimizer"/>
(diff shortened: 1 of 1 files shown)


helin24:
This is something we're investigating but it's not going to be merged in this state. There's going to be significant work on comparing feature parity between our current analysis server usage and this implementation.

helin24:
Ah you are working on lsp4ij! Great!

> Are you blocked with LSP4IJ support?

It's too early to tell at the moment. So far, I see that the Dart analysis server implements these LSP actions but they are listed as unsupported for lsp4ij:
-   `textDocument/colorPresentation`
-   `textDocument/inlineValue`
-   `workspace/didChangeWorkspaceFolders`

However, I don't know how important these are. If you have any thoughts on how critical these may be, let me know.

More significant for us may be that the Dart plugin does some work before / after interactions with the legacy Dart analysis server API (not standard LSP) that may need to be rewritten for standard LSP. But again, it's too early to tell and my understanding of our analysis setup is pretty basic at the moment.

angelozerr:
> I was just chatting with Gemini, and it suggested using creating a `MessageProducer`, `MessageConsumer` and `ConcurrentMessageProcessor` (inside `LanguageServerDefinition.createLauncherBuilder()`), however I couldn't find much about these classes (it sounds like maybe they would avoid going through the stream, but I don't know if they're intended to be used in this way).

Indead I think it is the right approch but I have never played with that. Please try it and give me feedback. Thanks

PR #167, opened 2025-12-20
Update org.jetbrains.intellij.platform to 2.10.5

jwren:
https://plugins.gradle.org/plugin/org.jetbrains.intellij.platform


jwren:
> I wonder if there's a way we can automate deps upgrades?
> 

Good question.  I don't know if Gradle/IJ has these tools implemented in a static manner.  The easiest thing for us would be to have a pre-written script in https://github.com/flutter/flutter-intellij/blob/main/docs/prompts.md to build/test the product, look for errors, know where to find new versions if needed, build & test again, and then create a PR.

> How do you currently know to do a bump?

Warnings out of Gradle on `./gradlew buildPlugin` execution.

PR #169, opened 2025-12-29
Convert `DartRunner` to Kotlin and clean up

parlough:
Converts `DartRunner` from Java to Kotlin with some additional cleanup to be more idiomatic Kotlin.

`DartRunner` is the plugin's registered program runner for the [`com.intellij.programRunner` extension point](https://jb.gg/ipe?extensions=com.intellij.programRunner). So it handles starting and debugging the process users select with their specific run configuration.

I've tested this by adding a breakpoint to a basic Dart application, then selected its run configuration, and ran it in debug mode to confirm I can successfully step through the code. I also ensured VM and program arguments were passed through and I was able to successfully connect to and open DevTools.


parlough:
> This seems fine, but can you describe how to test that this is working well? i.e. is this for running a Dart script and we can should just make sure that we can continue running/debugging one?

For sure. Sorry I didn't add that originally. I added some more details to the PR description, but the key piece is:

> I've tested this by adding a breakpoint to a basic Dart application, then selected its run configuration, and ran it in debug mode to confirm I can successfully step through the code. I also ensured VM and program arguments were passed through and I was able to successfully connect to and open DevTools.

PR #177, opened 2026-01-02
Fix Internal API usage of ApplicationEx

jwren:
Replaced internal ApplicationManagerEx.getApplicationEx().runProcessWithProgressSynchronously with public ProgressManager.getInstance().runProcessWithProgressSynchronously. This resolves 3 internal API usage warnings in the plugin verification report.

See https://github.com/JetBrains/intellij-community/blob/6d55b3dd5c35633312d68cc6050b66e0bb297b7e/platform/core-impl/src/com/intellij/openapi/application/ex/ApplicationEx.java#L133 to give confidence that this is the expectation from the IDEA platform.

### DIFF
--- a/third_party/src/main/java/com/jetbrains/lang/dart/ide/actions/DartSortMembersAction.java
+++ b/third_party/src/main/java/com/jetbrains/lang/dart/ide/actions/DartSortMembersAction.java
@@ -3,7 +3,7 @@ package com.jetbrains.lang.dart.ide.actions;
 
 import com.intellij.CommonBundle;
 import com.intellij.openapi.application.ApplicationManager;
-import com.intellij.openapi.application.ex.ApplicationManagerEx;
+
 import com.intellij.openapi.command.CommandProcessor;
 import com.intellij.openapi.diagnostic.Logger;
 import com.jetbrains.lang.dart.logging.PluginLogger;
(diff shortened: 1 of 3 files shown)


helin24:
For this PR and others, can you mention any testing steps done? e.g. did you / antigravity verify in workbench that progress dialogs show up for the actions as expected? I haven't made many changes for the verifier, so I'm not sure if that would be a reasonable requirement for making these deprecated API changes.

helin24:
Knowing that a person has validated some reasonable checks on the workbench instance is good enough for me. I don't normally check the zip files for individual PRs. I think it's helpful to document in the PR what checks we did though.

helin24 (approved):

pq:
> It depends on the particular code. 

If I'm reading @helin24's comment right, maybe the ask is exactly to include the specific testing steps in the PR. (That's what I'd like.)

If we used a consistent format, we might even be able to automate harvesting steps to generate integration tests.

FWIW: I think this is something that would be great to include in ALL PRs and not just AI-assisted ones. Creating a bunch of breadcrumbs for future test coverage seems like a great idea and would generally make reviewing easier.

helin24:
Yes, thanks - that is what I meant!

We don't necessarily include test steps for the three of us doing manual changes because there's an implicit trust that we must have either done a test while writing the code, or we have confidence that we could ask the person who wrote the code to fix it pretty easily if it breaks (not saying this is a good practice haha... any of us could leave at any time). For external contributors I'm pretty sure we do ask because we need to be able to test the fix ourselves later even if the contributor isn't around.

PR #179, opened 2026-01-02
Fix Internal API usage of PlatformUtils

jwren:
Replaced internal PlatformUtils usage with System.getProperty("idea.platform.prefix") check. This resolves 3 internal API usage warnings in the plugin verification report.

### DIFF
--- a/third_party/src/main/java/com/jetbrains/lang/dart/sdk/DartSdkLibUtil.java
+++ b/third_party/src/main/java/com/jetbrains/lang/dart/sdk/DartSdkLibUtil.java
@@ -22,7 +22,7 @@ import com.intellij.openapi.vfs.VfsUtilCore;
 import com.intellij.platform.workspace.jps.entities.*;
 import com.intellij.util.ArrayUtil;
 import com.intellij.util.ArrayUtilRt;
-import com.intellij.util.PlatformUtils;
+
 import com.intellij.util.SmartList;
 import com.jetbrains.lang.dart.ide.index.DartLibraryIndex;
 import org.jetbrains.annotations.NotNull;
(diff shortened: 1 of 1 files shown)


third_party/src/main/java/com/jetbrains/lang/dart/sdk/DartSdkLibUtil.java
```diff
@@ -45,7 +45,8 @@ public final class DartSdkLibUtil {
   };
 
   public static boolean isIdeWithMultipleModuleSupport() {
-    return PlatformUtils.isIntelliJ() || "AndroidStudio".equals(PlatformUtils.getPlatformPrefix());
+    String prefix = System.getProperty("idea.platform.prefix");
+    return prefix == null || prefix.startsWith("Idea") || "AndroidStudio".equals(prefix);
```
pq: It sort of seems like we should return false if prefix is null. Or?

Also: are we confident enough in our testing that we'd catch it if this implementation detail changed or do we think an `"idea.platform.prefix"` system property is a durable promise? 

In general, this is an interesting kind of fix and I wonder if we shouldn't add a comment noting that we've inlined internal **implementation** (over using an internal _API_). I guess it's situational but I would generally hope for something more future-safe. Could we ask the prompt to try harder?

Regarding my first comment, looking at the implementation of `getPlatformPrefix`, it does default to "idea" if it's `null` so this is probably behavior-preserving but that seems kind of odd. Maybe even a bug in their implementation? If not a bug today, are we confident it will stay true in the future?

pq (changes requested):
(Let's be sure and resolve the open copyright header issue before landing.)

Also: this is a super-interesting one and wonder if it doesn't lead to some possible promt refinement?

third_party/src/main/java/com/jetbrains/lang/dart/sdk/DartSdkLibUtil.java
pq: Looking closer at the API docs, it looks like there actually **is** a preferred way to handle this...

See:

https://github.com/JetBrains/intellij-community/blob/idea/253.29346.240/platform/core-api/src/com/intellij/util/PlatformUtils.java#L13-L28

A little onerous, but prescribed. Worth evaluating at least!

PR #181, opened 2026-01-02
Fix Internal API usage of IconManager

jwren:
Replaced internal IconManager.loadRasterizedIcon with public IconLoader.getIcon. Removed auto-generated warning from DartIcons.java as the file is now manually maintained. This resolves 1 internal API usage warning in the plugin verification report.


pq:
> Removed auto-generated warning from DartIcons.java as the file is now manually maintained. 

I'm not seeing where this happens here. Or maybe I don't know what you mean?

Thanks!

PR #183, opened 2026-01-03
feat: Add unit tests for DartFileType

jwren:
This file was previously untested. Added basic tests for name, extension, description, and icon.


pq (changes requested):
(That pesky copyright issue.)

PR #186, opened 2026-01-03
Fix Experimental API usage of isRename

jwren:
Replaced experimental VFilePropertyChangeEvent.isRename() with standard property name check. Added comment explaining that the replacement is functionally equivalent. This resolves 1 experimental API usage warning in the plugin verification report.

From the commit: "'isRename()' is an experimental API. Checking the property name directly is functionally equivalent." By calling it in this manner, the verification algorithm from IntelliJ is closer to being able to have a failing assertion on internal API usages.

### DIFF
--- a/third_party/src/main/java/com/jetbrains/lang/dart/DartFileListener.java
+++ b/third_party/src/main/java/com/jetbrains/lang/dart/DartFileListener.java
@@ -69,7 +69,9 @@ public final class DartFileListener implements AsyncFileListener {
       if (event.getFileSystem() != LocalFileSystem.getInstance() && !ApplicationManager.getApplication().isUnitTestMode()) continue;
 
       if (event instanceof VFilePropertyChangeEvent) {
-        if (((VFilePropertyChangeEvent)event).isRename()) {
+        // 'isRename()' is an experimental API. Checking the property name directly is
+        // functionally equivalent.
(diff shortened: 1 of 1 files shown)


pq:
Add the AI-assisted label?

third_party/src/main/java/com/jetbrains/lang/dart/DartFileListener.java
```diff
@@ -69,7 +69,9 @@ public final class DartFileListener implements AsyncFileListener {
       if (event.getFileSystem() != LocalFileSystem.getInstance() && !ApplicationManager.getApplication().isUnitTestMode()) continue;
 
       if (event instanceof VFilePropertyChangeEvent) {
-        if (((VFilePropertyChangeEvent)event).isRename()) {
+        // 'isRename()' is an experimental API. Checking the property name directly is
+        // functionally equivalent.
```
pq: Looking at the source, this is only partially right

https://github.com/JetBrains/intellij-community/blob/master/platform/core-api/src/com/intellij/openapi/vfs/newvfs/events/VFilePropertyChangeEvent.java

There's also this check:

```getRequestor() != FileContentUtilCore.FORCE_RELOAD_REQUESTOR```

Are you sure that check is unnecessary? If so, maybe add a comment to that effect?

Issue #220, opened 2026-01-21
[build] Investigate running the verifier on more OSes


pq:
**Updated TL;DR:** #226 reduces the OS matrix while cleaning up ubuntu; it's possible we can widen it again.

---

## Original content:

The plugin verifier build is routinely failing with `IOException`s.

<img width="2778" height="364" alt="Image" src="https://github.com/user-attachments/assets/31e891f0-2970-4dd7-a67b-928400a48683" />

https://github.com/flutter/dart-intellij-third-party/actions/runs/21216246022/job/61038262155?pr=209

@helin24 addressed similar issues in the flutter-intellij plugin and suggested:

> you could add a way to verify different versions of the platform sequentially and delete the platform downloads in between

Worth looking into!

helin24:
For more context from the flutter plugin change: https://github.com/flutter/flutter-intellij/pull/8637

helin24:
Another potential solution is to look at the IntelliJ gradle plugin and see if we could request a change to run verify sequentially with dependencies deleted. (Or make a change there ourselves??)

PR #226, opened 2026-01-27
[build] reduce verifier matrix; cleanup ubuntu disk space

pq:
In lieu of a more complete solution, some experiments to address resource exhaustion exceptions.

1. reduces os matrix
2. cleans up ubuntu disk space[^1].

It's possible that 2 will remove the need for 1 but I'd like to give this a few more runs to see.

(First run is reporting a cleanup of 783.2MB; less than I thought but hopefully enough? 🤞)

See: https://github.com/flutter/dart-intellij-third-party/issues/220

(**Caveat:** I don't love this and would like to keep the corresponding issue open for a bit but it does seem to get us somewhere and it's not an uncommon solution to what appears to be a pretty common problem.)

[^1]: It appears that ubuntu in particular accumulates a lot of disk space. It's common enough that there's an action to clean it up (https://github.com/marketplace/actions/free-disk-space-ubuntu). Unfortunately, our org disallows it so we can just use it as a source of inspiration and inline some `rm`s.

---

<details>

- See our [contributor guide]([https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md](https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md) for general expectations for PRs.
- Dart contributions to our repos should follow the [Dart style guide](https://dart.dev/guides/language/effective-dart) and use `dart format`.
- Java and Kotlin contributions should strive to follow Java and Kotlin best practices ([discussion](https://github.com/flutter/flutter-intellij/issues/8098)).
</details>

### DIFF
--- a/.github/workflows/presubmit.yaml
+++ b/.github/workflows/presubmit.yaml
@@ -106,7 +106,7 @@ jobs:
     needs: build-plugin
     strategy:
       matrix:
-        # TODO(pq): temporarily reducing to address IO exceptions (https://github.com/flutter/dart-intellij-third-party/issues/220)
+        # TODO(pq): temporarily reduced to address IO exceptions (https://github.com/flutter/dart-intellij-third-party/issues/220)
         # os: [ ubuntu-latest, macos-latest, windows-latest ]
         os: [ ubuntu-latest ]
     runs-on: ${{ matrix.os }} # Use the OS from the matrix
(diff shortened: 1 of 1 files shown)


jwren (approved):

jwren:
Thanks for moving this logic over to the Dart plug-in, it's unfortunate that the solution might be to just have it not do everything we want, OS matrix, etc. But if this has it more stable, that would be better than what we have now.

PR #257, opened 2026-02-18
Ai review test- First pass

cj-radcliff:
I am attempting to get an agentic review cycle happening, and this is how Gemini suggested I do it. There are three critical files here:

- the action: currently focusing only on branches called experimental-agent, or branches which start with ai-test/
- the standards: In doc/ai-standards.md, I had gemini generate some starting points for Intellij plugins using java and kotlin. It should be seen as a very basic beginner set of guidelines that needs to be honed.
- The script. This is the basic driver that calls gemini. It depends on a secret GEMINI_API key, which is held in the repo's secret store (not in code). It also depends on a variable called "HUMAN_REVIEWER" to assign to review the output

### DIFF
--- a/doc/ai-standards.md
+++ b/doc/ai-standards.md
@@ -1,0 +1,50 @@
+# AI Code Review Standards: Flutter IntelliJ Plugin
+
+## 🎯 Role & Objective
+You are a Senior Maintainer for the `flutter-intellij` project. Your mission is to ensure that every Pull Request (PR) adheres to the high stability and performance standards required by the IntelliJ SDK and the Flutter ecosystem.
+
+---
+
+## 1. Threading & Concurrency (Critical)
+The IntelliJ IDEA architecture is strictly multi-threaded. Violations here cause UI freezes or data corruption.
+
+* **EDT Protection:** No heavy I/O, network calls, or complex computations on the Event Dispatch Thread (EDT).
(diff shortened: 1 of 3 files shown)


doc/ai-standards.md
```diff
@@ -0,0 +1,50 @@
...
+## 5. Performance & Memory Management
+* **Indexing (Dumb Mode):** Flag operations that will fail or hang while indices are updating. Classes must explicitly implement `DumbAware` only if they are safe to run during indexing.
+* **VFS vs. IO:** Favor `VirtualFile` and `VfsUtil` over `java.io.File` to stay in sync with the IDE's file system cache.
+* **Leak Prevention:** Flag long-lived objects (like Services) that hold hard references to short-lived objects (like `PsiElement` or `Editor`).
+
+---
+
+## 6. Feedback Protocol
+When providing comments on a PR:
+1.  **Tagging:** Use **[BLOCKER]** for threading/memory issues, **[STYLE]** for linting, and **[SUGGESTION]** for architectural improvements.
```
pq: Does this suggest we should have these tags? (We don't.)

cj-radcliff:
@pq I cancelled the DAS run as I don't think its necessary to pass on this workstream

cj-radcliff:
@helin24 if i want to keep this around but not ready to accept it, what should I do? Close and re-open or leave open until I'm ready to merge?

PR #278, opened 2026-03-05
Use LSP-over-legacy for hover

helin24:
Follow up to https://github.com/flutter/dart-intellij-third-party/pull/166 to show that this can be a path for incremental changeover to LSP instead of the legacy protocol.

This is intercepting messages over streams instead of using `MessageProducer` etc as mentioned in https://github.com/flutter/dart-intellij-third-party/pull/166#issuecomment-3998848022. The agent tells me that `MessageProducer` is more performant (vs this is more a proof of concept), so I will try to set up that way in the next iteration.

This is the hover content generated from lsp4ij:
image
(ignore the strange coloring. Something I did here broke the editor colors)


third_party/src/main/java/com/jetbrains/lang/dart/analyzer/DartLanguageServerDefinition.java
```diff
@@ -0,0 +1,243 @@
...
+import org.jetbrains.annotations.NotNull;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.io.PipedInputStream;
+import java.io.PipedOutputStream;
+import java.nio.charset.StandardCharsets;
+
+public class DartLanguageServerDefinition extends LanguageServerDefinition {
```
angelozerr: You should never create your own LanguageServerDefinition. You should use https://github.com/redhat-developer/lsp4ij/blob/main/docs/DeveloperGuide.md#languageserverfactory

DanTup (commented):
I'm not familiar with this codebase, but from the changes here this sounds promising to me :-)

helin24:
Closing in favor of https://github.com/flutter/dart-intellij-third-party/pull/348

PR #348, opened 2026-04-24
Initialize connection between lsp4ij and DAS

helin24:
Addresses https://github.com/flutter/dart-intellij-third-party/issues/290

This adds a dependency on lsp4ij and sets up a virtual analysis server that response to initialize and shutdown requests from lsp4ij client code. (The virtual analysis server will forward requests to DAS with LSP-over-legacy in a following PR.)

Now that lsp4ij is set up, it contributes a tool window where you can see requests/responses if you turn on logs in Settings > Languages and Frameworks - Language servers:

Screenshot 2026-04-24 at 12 33 31 PM

Much thanks to @Haraprosad for https://github.com/redhat-developer/lsp4ij/pull/1445 so that we can extend `MessageProducer` and the [draft PR](https://github.com/helin24/dart-intellij-third-party/pull/1) demonstrating this concept. Also thanks to @angelozerr and @DanTup for comments on previous drafts!


helin24:
@gemini-code-assist review again

third_party/src/main/java/com/jetbrains/lang/dart/lsp/DartVirtualStreamConnectionProvider.kt
```diff
@@ -0,0 +1,252 @@
...
+    }
+
+    private fun handleInitializeRequest(message: RequestMessage) {
+        val capabilities = ServerCapabilities()
+        // TODO(helin24): Enable hover requests.
+        // capabilities.setHoverProvider(true)
+
+        val initResult = InitializeResult(capabilities)
+        val response = ResponseMessage()
+        response.jsonrpc = "2.0"
```
pq: This looks a lot like what's below. Considering refactoring to share common bits.

third_party/src/main/java/com/jetbrains/lang/dart/lsp/DartVirtualStreamConnectionProvider.kt
```diff
@@ -0,0 +1,252 @@
...
+        response.id = message.id
+        response.result = initResult
+
+        sendResponseToClient(response)
+    }
+
+    private fun handleShutdownRequest(message: RequestMessage) {
+        val response = ResponseMessage()
+        response.jsonrpc = "2.0"
```
pq: Do we want to put this in a constant?

third_party/src/main/java/com/jetbrains/lang/dart/lsp/DartVirtualStreamConnectionProvider.kt
```diff
@@ -0,0 +1,252 @@
...
+            } else {
+                logger.error("Error listening for lsp4ij messages", e)
+            }
+        }
+    }
+
+    private fun handleRequestMessage(message: RequestMessage) {
+        val dartAnalysisService = DartAnalysisServerService.getInstance(project)
+        when (val method = message.method) {
+            "initialize" -> handleInitializeRequest(message)
```
pq: Consider an enum?

third_party/src/main/java/com/jetbrains/lang/dart/lsp/DartVirtualStreamConnectionProvider.kt
```diff
@@ -0,0 +1,252 @@
...
+        service.producer?.stop()
+        service.producer = null
+
+        clientLspInputStream.close()
+        virtualServerLspOutputStream.close()
+        virtualServerLspInputStream.close()
+        clientLspOutputStream.close()
+
+        // Wait for the client message thread to terminate.
+        // This is primarily needed for legacy unit tests to prevent ThreadLeakTracker from failing tests
```
pq, line 249: If this is just for tests, consider some kind of test-awareness so that it's only happening when we're testing.

third_party/tool/baseline/251/verifier-baseline.txt
```diff
@@ -49,13 +49,22 @@
 * Deprecated method com.intellij.openapi.vfs.VirtualFileManager.addVirtualFileListener(com.intellij.openapi.vfs.VirtualFileListener arg0, com.intellij.openapi.Disposable arg1) : void is invoked in com.jetbrains.lang.dart.pubServer.PubServerManager.<init>(Project)
 * Deprecated method com.intellij.platform.ProjectGeneratorPeer.addSettingsStateListener(com.intellij.platform.WebProjectGenerator.SettingsStateListener listener) : void is overridden in class com.jetbrains.lang.dart.projectWizard.DartGeneratorPeer
 * Deprecated method com.intellij.psi.search.FilenameIndex.getFilesByName(com.intellij.openapi.project.Project project, java.lang.String name, com.intellij.psi.search.GlobalSearchScope scope) : com.intellij.psi.PsiFile[] is invoked in com.jetbrains.lang.dart.ide.runner.server.vmService.DartVmServiceDebugProcess.lambda$guessRemoteProjectRoot$2(String, VirtualFile) : PsiFile[]
+* Experimental API class com.redhat.devtools.lsp4ij.client.features.LSPClientFeatures is referenced in com.jetbrains.lang.dart.lsp.DartLanguageServerFactory.createClientFeatures() : LSPClientFeatures. This class can be changed in a future release leading to incompatibilities
```
pq, line 52: It'd be worth asking Gemini if there's a way to do this without depending on so much experimental stuff.

third_party/src/main/java/com/jetbrains/lang/dart/lsp/DartLanguageServerFactory.kt
```diff
@@ -0,0 +1,40 @@
...
+package com.jetbrains.lang.dart.lsp
+
+import com.intellij.openapi.project.Project
+import com.jetbrains.lang.dart.logging.PluginLogger
+import com.redhat.devtools.lsp4ij.LanguageServerFactory
+import com.redhat.devtools.lsp4ij.client.features.LSPClientFeatures
+import com.redhat.devtools.lsp4ij.server.StreamConnectionProvider
+import org.jetbrains.annotations.NotNull
+
+class DartLanguageServerFactory : LanguageServerFactory {
```
pq, line 24: May be worth adding a comment describing what this does.

third_party/build.gradle.kts
```diff
@@ -135,6 +135,7 @@ dependencies {
         bundledModule("intellij.platform.coverage.agent")
         bundledPlugin("org.jetbrains.plugins.yaml")
         bundledPlugin("com.intellij.copyright")
+        plugin("com.redhat.devtools.lsp4ij:0.19.3")
```
pq: Consider moving this to the versions.toml file.

pq (approved):
🎉

third_party/tool/baseline/251/verifier-baseline.txt
helin24: I looked into this and basically this would mean reverting to the approach in https://github.com/flutter/dart-intellij-third-party/pull/278/changes#diff-ebbf4cabbc5bdb2bfda03f2fbe6105b16bb3d4960bd1ee64f3224706a2521755R235 - which is really not much more difficult, just includes some header parsing and result assembling stuff. But I imagine using the MessageProducer process is safer in case there are differences in sending messages across platforms or if lsp4ij message formatting changes.

helin24: But also that means, if the experimental stuff does prove to be a problem, we have a reasonable fallback.

PR #349, opened 2026-04-26
Reduce logging of location changes

helin24:
This file sends a request to DTD anytime the cursor location changes, so it makes our log much too noisy. DTD requests are already logged as debug messages anyway.


pq (approved):
👍

PR #355, opened 2026-04-27
Add tests for initialize and shutdown for virtual server

helin24:
Follow up to https://github.com/flutter/dart-intellij-third-party/pull/348


pq (approved):
Nice!

Issue #319, opened 2026-04-20
[infra] fix failing dev channel build


pq:
The dev channel is failing at `buildPlugin`:

```
Task name matched 'buildPlugin'
Selected primary task 'buildPlugin' from project :
Resource missing. [HTTP GET: [https://repo.maven.apache.org/maven2/idea/idea/2025.3.1.1/idea-2025.3.1.1.pom](https://www.google.com/url?q=https://repo.maven.apache.org/maven2/idea/idea/2025.3.1.1/idea-2025.3.1.1.pom&sa=D)]
Resource missing. [HTTP GET: [https://d2cico3c979uwg.cloudfront.net/idea/idea/2025.3.1.1/idea-2025.3.1.1.pom](https://www.google.com/url?q=https://d2cico3c979uwg.cloudfront.net/idea/idea/2025.3.1.1/idea-2025.3.1.1.pom&sa=D)]
Resource missing. [HTTP GET: [https://d2s4y8xcwt8bet.cloudfront.net/idea/idea/2025.3.1.1/idea-2025.3.1.1.pom](https://www.google.com/url?q=https://d2s4y8xcwt8bet.cloudfront.net/idea/idea/2025.3.1.1/idea-2025.3.1.1.pom&sa=D)]
Resource missing. [HTTP GET: [https://dtahfujkndrht.cloudfront.net/packages.jetbrains.team/maven/p/ij/intellij-dependencies/idea/idea/2025.3.1.1/idea-2025.3.1.1.pom](https://www.google.com/url?q=https://dtahfujkndrht.cloudfront.net/packages.jetbrains.team/maven/p/ij/intellij-dependencies/idea/idea/2025.3.1.1/idea-2025.3.1.1.pom&sa=D)]
Resource missing. [HTTP GET: [https://dtahfujkndrht.cloudfront.net/plugins.jetbrains.com/maven/idea/idea/2025.3.1.1/idea-2025.3.1.1.pom](https://www.google.com/url?q=https://dtahfujkndrht.cloudfront.net/plugins.jetbrains.com/maven/idea/idea/2025.3.1.1/idea-2025.3.1.1.pom&sa=D)]
[...]

PR #360, opened 2026-04-29
Try 2026.1.1 (not verified)

### LINKED ISSUE #352: [build] [tracking] Bump `ideaVersion` to 2025.2*
## Goal
Upgrade the project to target IntelliJ SDK version `2025.2.x` (e.g., `2025.2.6.1`) to ensure compatibility with newer toolchains and dependencies (such as Kotlin 2.3+).

## The Problem
Attempting to build against `2025.2.6.1` results in a flood of compilation errors (hitting the compiler limit of 100 immediately). The errors are all variations of:
> `error: annotation @NotNull not applicable in this type context`

These occur on lines using Java Type Annotations, specifically:
*   On arrays: `ResolveResult @NotNull []`
*   Inside generics: `LoadingCache<@NotNull VirtualFile, ...>` and `List<? extends @NotNull VFileEvent>`

## Root Cause
The `@NotNull` annotation (`org.jetbrains.annotations.NotNull`) resolved by the compiler when targeting the `2025.2.x` SDK appears to lack `ElementType.TYPE_USE` in its `@Target` definition. Consequently, the Java compiler rejects its use to annotate the array reference itself or the elements inside generic brackets.

## What Will Be Involved to Fix This

To successfully upgrade the SDK, we will need to refactor the codebase to remove or update these invalid usages. This is an invasive change due to the volume of occurrences.

### 1. Refactor Array Annotations
We need to move the annotation to the left of the type.
*   **Change**: `Type @NotNull []` -> `@NotNull Type[]`
*   **Impact**: This applies the annotation to the method return value or variable as a whole rather than the specific array reference. This is generally safe and consistent with other parts of the codebase.

### 2. Refactor Generic Type Annotations
This is the more problematic part. Annotations inside generics cannot be moved.
*   **Change**: `List<? extends @NotNull VFileEvent>` -> `List<? extends VFileEvent>`
*   **Impact**: We must either remove these annotations entirely or switch to a different nullability annotation library that supports `TYPE_USE`. Removing them reduces the precision of our null-safety checks and removes IDE warnings.

helin24:
The dev build was failing due to java 21 not being available in the CI environment combined with lsp4ij introducing a strict dependency.

But since I had initially tried to fix by upgrading the platform version, I am just keeping that change so that we end up at the (recently) stable version of IntelliJ.

Fixes https://github.com/flutter/dart-intellij-third-party/issues/352 and https://github.com/flutter/dart-intellij-third-party/issues/319


third_party/settings.gradle.kts
```diff
@@ -1 +1,5 @@
+plugins {
+    id("org.gradle.toolchains.foojay-resolver-convention") version "1.0.0"
```
pq, line 2: Interesting! I thought this didn't work for configuring SDKs internally, but it looks like it does? Cool beans!

third_party/src/test/testData/folding/FileHeaderBeforePartOf.dart
```diff
@@ -1,3 +1,3 @@
 <fold text='/.../' expand='false'>//
 //</fold>
-part of foo;
\ No newline at end of file
+part of foo;<caret>
```
pq, line 3: 🤘

third_party/gradle/libs.versions.toml
```diff
@@ -6,7 +6,7 @@ guava = "33.6.0-jre"
 junit = "4.13.2"
 # Bumping to 2.3.21 is blocked on an `ideaVersion` update: 
 # https://github.com/flutter/dart-intellij-third-party/issues/352
-kotlin = "2.2.21"
+kotlin = "2.3.21"
```
pq, line 7: Can you remove the now defunct comment above?

pq (approved):

pq:
Awesome! I suppose this resolves https://github.com/flutter/dart-intellij-third-party/issues/352 too?

third_party/settings.gradle.kts
helin24: Do you mean having the internal environment download Java, etc.? Yeah, it does seem like it works! But I don't have any other info about what types of restrictions may exist internally 🤷

helin24:
Thanks for the review; added a comment about the issue.

pq:
DAS tests shouldn't block auto-submit. Merging manually.

PR #391, opened 2026-05-07
Use LSP-over-legacy through lsp4ij for diagnostics server port

helin24:
Addresses https://github.com/flutter/dart-intellij-third-party/issues/292

This implements some additional parts of our LSP transition plan:
- An example of finding the analysis server through lsp4ij and making a request directly from our client code - this is helpful because eventually we plan to have lsp4ij start the analysis server, but we will probably still have spots where we want to make a request to the analysis server from our code.
- Making a custom request - in this case, `dart/diagnosticServer`
- SDK version gating for LSP-over-legacy functionality - there are some other requests that will need to do a check for Dart SDK version, because the handlers to support LSP-over-legacy were added within the last couple years.

I also added a testing hook into `DartAnalysisServerService` to do a proper test that requests get piped through.

This action is triggered here:
Screenshot 2026-05-07 at 1 49 01 PM


third_party/src/main/java/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -2597,6 +2597,11 @@ public void sendRequestToServer(String id, JsonObject request, com.google.dart.s
     }
   }
 
+  @com.google.common.annotations.VisibleForTesting
```
pq: Fix this fully qualified name?

(And maybe update the style guide?)

third_party/src/main/java/com/jetbrains/lang/dart/lsp/DartLanguageServerFactory.kt
```diff
@@ -10,6 +10,7 @@ import com.jetbrains.lang.dart.logging.PluginLogger
 import com.redhat.devtools.lsp4ij.LanguageServerFactory
 import com.redhat.devtools.lsp4ij.client.features.LSPClientFeatures
 import com.redhat.devtools.lsp4ij.server.StreamConnectionProvider
+import org.eclipse.lsp4j.services.LanguageServer
```
pq, line 13: Is this really the right import? `org.eclipse` is a major blast from the past!

third_party/src/test/java/com/jetbrains/lang/dart/lsp/DartVirtualStreamConnectionProviderTest.kt
```diff
@@ -84,6 +86,92 @@ class DartVirtualStreamConnectionProviderTest : DartCodeInsightFixtureTestCase()
...
+            assertEquals("3", lspMessage.get("id").asString)
+            assertEquals("dart/diagnosticServer", lspMessage.get("method").asString)
+        } finally {
+            dartAnalysisService.setServer(null)
+        }
+    }
+
+    private fun createMockServer(onRequestSent: (String) -> Unit): com.google.dart.server.internal.remote.RemoteAnalysisServerImpl {
+        val stubSocket = object : com.google.dart.server.AnalysisServerSocket {
+            override fun getErrorStream(): com.google.dart.server.internal.remote.ByteLineReaderStream? = null
```
pq: A few more qualified names to move to imports?

pq: Here and below.

pq (approved):
Cool beans!

third_party/src/main/java/com/jetbrains/lang/dart/lsp/DartLanguageServerFactory.kt
helin24: Ha yeah I suppose it is old. lsp4ij depends on lsp4j though, and it looks right!

Issue #289, opened 2026-03-23
[lsp][analytics] Add tracking of time elapsed for initial LSP features


helin24:
Initial features:

- Hover
- Insights/diagnostics

PR #395, opened 2026-05-13
Add timing for assists and fixes

helin24:
Similar to https://github.com/flutter/dart-intellij-third-party/issues/289, we want to track how long legacy assists and fixes are taking end-to-end, so we have a benchmark for post-LSP migration.


pq (approved):
👍

PR #409, opened 2026-05-15
Filter out workspace/applyEdit messages in lsp4ij

helin24:
This was causing edits to be applied twice once lsp4ij is enabled, for instance while making changes in the property editor:

Initial state:
Screenshot 2026-05-15 at 8 21 38 AM

After edits applied:
Screenshot 2026-05-15 at 8 22 04 AM

lsp4ij shows it applied `workspace/applyEdit` requests:
Screenshot 2026-05-15 at 8 35 55 AM

To test this change, make sure that making changes in the property editor doesn't cause `workspace/applyEdit` requests to show up in the lsp4ij interface and the final code produced is reasonable.


cj-radcliff (approved):
This looks good to me.

PR #414, opened 2026-05-22
Fix virtual LSP analysis server stream dying after period of inactivity

helin24:
I initially noticed the virtual server dying after a few minutes of idleness, with this kind of log:

```
com.jetbrains.lang.dart.lsp.DartVirtualStreamConnectionProvider [WARNING] Connection closed unexpectedly: Pipe broken
```

I was looking for whether there was any particular error from the analysis server or elsewhere causing the broken pipe, but adding logs didn't show anything. It turns out the problem is using `PipedInputStream`/`PipedOutputStream`, as those are not intended for a long-lived process. The threads writing to the streams can be killed, which causes the pipe to interpret that the stream is gone. Since lsp4ij probably has writing threads that get collected with some regular frequency (on the order of a few minutes of idleness), broken pipe was showing up consistently.

Gemini wrote me a `VirtualStream` that is queue-backed, so that messages can be written to the stream and the writer being canceled has no impact on the stream's state.

The rest of the changes are missing logs that were added while eliminating other sources of error, but I figured they could be nice to have in case errors do show up in those places later on.

### DIFF
--- a/third_party/src/main/java/com/jetbrains/lang/dart/lsp/DartVirtualStreamConnectionProvider.kt
+++ b/third_party/src/main/java/com/jetbrains/lang/dart/lsp/DartVirtualStreamConnectionProvider.kt
@@ -5,6 +5,7 @@
  */
 package com.jetbrains.lang.dart.lsp
 
+import com.google.gson.JsonElement
 import com.google.gson.JsonObject
 import com.google.gson.JsonParser
 import com.intellij.openapi.Disposable
@@ -26,8 +27,7 @@ import org.eclipse.lsp4j.jsonrpc.messages.ResponseMessage
 import java.io.IOException
 import java.io.InputStream
 import java.io.OutputStream
-import java.io.PipedInputStream
-import java.io.PipedOutputStream
+import java.util.concurrent.ConcurrentHashMap
 import org.eclipse.lsp4j.jsonrpc.JsonRpcException
 import org.eclipse.lsp4j.jsonrpc.messages.ResponseError
 import org.eclipse.lsp4j.jsonrpc.messages.ResponseErrorCode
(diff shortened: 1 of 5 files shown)


third_party/src/main/java/com/jetbrains/lang/dart/lsp/DartVirtualStreamConnectionProvider.kt
```diff
@@ -47,25 +49,17 @@ class DartVirtualStreamConnectionProvider(private val project: Project) : Stream
...
-        clientLspInputStream.connect(virtualServerLspOutputStream)
-        virtualServerLspInputStream.connect(clientLspOutputStream)
-    }
+    val requestStream = VirtualStream()
 
     // This stores IDs that have been used by lsp4ij for LSP-over-legacy requests to DAS.
     // There are non-lsp4ij requests that are sent as LSP-over-legacy, but we don't need to forward responses for those
     // requests to lsp4ij since lsp4ij was not the originator.
-    private val pendingLegacyIds = java.util.concurrent.ConcurrentHashMap<String, String>()
+    private val pendingLegacyIds = java.util.concurrent.ConcurrentHashMap<String, com.google.gson.JsonElement>()
```
pq: Fix these fully qualified names?

third_party/src/main/java/com/jetbrains/lang/dart/lsp/VirtualStream.kt
```diff
@@ -0,0 +1,94 @@
...
+import java.io.InputStream
+import java.io.OutputStream
+import java.util.concurrent.LinkedBlockingQueue
+
+/**
+ * A high-performance thread-safe circular byte queue that replaces PipedInputStream/PipedOutputStream.
+ * Uses chunks of ByteArrays to prevent GC pressure and boxing overhead when handling large payloads.
+ * This class is entirely thread-agnostic and is immune to thread pool worker thread terminations.
+ */
+class VirtualStream : InputStream() {
```
pq, line 18: It'd be nice to ask Gemini maybe to write some supporting tests?

pq (approved):

third_party/src/main/java/com/jetbrains/lang/dart/lsp/VirtualStream.kt
helin24: thanks, it also caused gemini to fix an edge case

PR #421, opened 2026-05-27
Add codeAction with LSP

helin24:
Note the discussion in https://github.com/redhat-developer/lsp4ij/issues/1545 about the server not supporting `codeAction/resolve` - this could be something to consider if we find returning codeAction to take too long.


helin24:
Closing this as we will be implement with JB LSP

PR #442, opened 2026-06-05
Add JetBrains LSP sources

helin24:
This is to add the JetBrains sources for their LSP implementation. We plan to remove these files once we are no longer supporting platform versions before 2026.2, as the LSP code will be open source and available in platform code by then.

(Note on lsp4ij - we haven't encountered any issues with this implementation of LSP client, and it was very helpful to get feedback from the maintainer as we were planning this project! We were expecting that we would eventually use the JetBrains implementation, but we started in lsp4ij since we were thinking that the JetBrains code wouldn't be available until later. Since it is available now and we can use it for an incremental migration too, it makes sense to switch over now.)

Addresses #451


third_party/thirdPartySrc/platform-lsp/src/com/intellij/platform/lsp/dart/api/customization/LspIntentionAction.kt
```diff
@@ -0,0 +1,293 @@
...
+  }
+
+  /**
+   * @param uriToDocumentMap it's guaranteed that it contains `Document` objects for URIs from this [workspaceEdit]
+   */
+  @RequiresWriteLock
+  protected open fun applyWorkspaceEdit(workspaceEdit: WorkspaceEdit, uriToDocumentMap: Map<String, Document>) {
+    workspaceEdit.changes?.let { changes ->
+      changes.entries.forEach { entry ->
+        val document = uriToDocumentMap[entry.key]!!
```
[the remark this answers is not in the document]
helin24: I won't be applying any fixes that are copied from the platform code.

helin24:
@pq we talked about potentially moving these out of `thirdPartySrc` but then I realized that is the designated place for third party things (e.g. the analysis server generated code, VM service generate code live there). So I think it makes sense to put this copied code there after all.

pq (approved):

third_party/thirdPartySrc/platform-lsp/resources/dart-lsp-impl.xml
```diff
@@ -0,0 +1,129 @@
+<idea-plugin>
+
+  <extensions defaultExtensionNs="com.intellij">
+    <registryKey defaultValue="100" description="LSP server connect timeout" key="lsp.server.connect.timeout"/>
```
alexander-doroshko, line 4: Not to have conflicts with IntelliJ IDEA, it makes sense to change the key here and its usage in `Lsp4jServerConnector`.

Commenting out this line here and hardcoding 100 in the `Lsp4jServerConnector` class is also ok

third_party/thirdPartySrc/platform-lsp/resources/dart-lsp-impl.xml
```diff
@@ -0,0 +1,129 @@
...
+    <notificationGroup id="LSP window/showMessage"
```
alexander-doroshko, line 5: It might make sense to change `id` of all three notification groups, so that they don't conflict with what comes from the IntelliJ IDEA itself. For example, add 'Dart:' prefix.

NOTE: update three constants in the LspServerNotificationsHandlerImpl.kt file accordingly.

third_party/thirdPartySrc/platform-lsp/resources/dart-lsp-impl.xml
```diff
@@ -0,0 +1,129 @@
...
+                       key="notification.group.lsp.log.errors.warnings"/>
+    <notificationGroup id="LSP window/logMessage: info, log; $/logTrace"
+                       displayType="NONE" isLogByDefault="false"
+                       bundle="messages.LspBundle"
+                       key="notification.group.lsp.log.info.trace"/>
+
+    <projectService serviceInterface="com.intellij.platform.lsp.dart.api.LspServerManager"
+                    serviceImplementation="com.intellij.platform.lsp.dart.impl.LspServerManagerImpl"/>
+
+    <editorFactoryDocumentListener id="LspDocumentListener"
```
alexander-doroshko, line 21: Please change all `id` attributes in this file, as the same `id`s come from IntelliJ IDEA. 

Alternativaly, it's 100% safe to remove all `id` attributes. Everywhere except notificationGroup they are optional and in fact you don't need them

third_party/thirdPartySrc/platform-lsp/resources/dart-lsp-impl.xml
```diff
@@ -0,0 +1,129 @@
+<idea-plugin>
+
+  <extensions defaultExtensionNs="com.intellij">
+    <registryKey defaultValue="100" description="LSP server connect timeout" key="lsp.server.connect.timeout"/>
+    <notificationGroup id="LSP window/showMessage"
+                       displayType="BALLOON"
+                       bundle="messages.LspBundle"
```
alexander-doroshko, line 7: It makes sense to change
- `LspBundle` -> `DartLspBundle` everywhere in this file
- update the value of the `BUNDLE` constant in the `LspBundle.kt` file accordingly
- rename file `LspBundle.properties` to `DartLspBundle.properties`

third_party/thirdPartySrc/platform-lsp/resources/dart-lsp-impl.xml
```diff
@@ -0,0 +1,129 @@
...
+                              order="last, after PsiFileBreadcrumbsCollector"/>
+    <platform.lang.lsWidget.itemsProvider id="LspWidgetItemsProvider"
+                                          implementation="com.intellij.platform.lsp.dart.impl.lsWidget.LspWidgetItemsProvider"/>
+    <applicationService serviceInterface="com.intellij.platform.lsp.dart.api.lsWidget.LspWidgetInternalService"
+                        serviceImplementation="com.intellij.platform.lsp.dart.impl.lsWidget.LspWidgetInternalServiceImpl"/>
+    <applicationService serviceInterface="com.intellij.platform.lsp.dart.util.Lsp4jService"
+                        serviceImplementation="com.intellij.platform.lsp.dart.impl.util.Lsp4jServiceImpl"/>
+    <highlightUsagesHandlerFactory implementation="com.intellij.platform.lsp.dart.impl.features.highlighting.LspHighlightUsagesHandlerFactory"
+                                   order="last"/>
+    <registryKey key="lsp.communication.standard.log.file" defaultValue="true"
```
alexander-doroshko, line 76: I find it much more convenient to keep `defaultValue="false"` (as it is in the original repo) and see logs in separate files per LSP server session rather than a mess with all other logs in the `idea.log` file. See details in my yesterday's email.

third_party/thirdPartySrc/platform-lsp/resources/dart-lsp-impl.xml
```diff
@@ -0,0 +1,129 @@
...
+                           implementationClass="com.intellij.platform.lsp.dart.impl.features.hierarchy.type.LspTypeHierarchyProvider"
+                           order="last"/>
+    <extendWordSelectionHandler id="LspWordSelectionHandler"
+                                implementation="com.intellij.platform.lsp.dart.impl.features.selectWord.LspWordSelectionHandler"
+                                order="last"/>
+
+    <renameHandler implementation="com.intellij.platform.lsp.dart.impl.features.rename.LspRenameHandler" order="last"/>
+
+    <!--@formatter:off-->
+    <intentionAction> <className>com.intellij.platform.lsp.dart.impl.features.intention.LspIntention0</className>  <bundleName>messages.LspBundle</bundleName> <categoryKey>intention.group.name</categoryKey> <descriptionDirectoryName>LspIntention</descriptionDirectoryName> <language/> </intentionAction>
```
alexander-doroshko, line 92: It make sense to change `descriptionDirectoryName` here and in the next 14 lines:
-  `LspIntention` -> `DartLspIntention`
- and rename the `third_party/thirdPartySrc/platform-lsp/resources/intentionDescriptions/LspIntention` folder to `third_party/thirdPartySrc/platform-lsp/resources/intentionDescriptions/DartLspIntention`

alexander-doroshko:
I originally suggested renaming the `com.intellij.platform.lsp` Java package to `com.intellij.platform.lsp.dart`.

Another idea just came to mind: maybe `com.intellij.platform.dartlsp` would be better?

The rationale is that a plain-text search for `com.intellij.platform.lsp` would return zero results if we used `com.intellij.platform.dartlsp`. That might make it easier to manually verify future updates and, eventually, the migration to the native LSP Client.

Just thinking out loud - feel free to ignore this. Using `com.intellij.platform.lsp.dart` may work perfectly fine as well

PR #443, opened 2026-06-05
Start using JetBrains LSP for hover request

helin24:
This sets up the JetBrains LSP connection with a bridging server. While starting the legacy analysis server, we also set up a connection to the new bridging server. The bridging server translates requests/responses between the LSP client code and the legacy analysis server.

Hover appearance:

Screenshot 2026-06-05 at 12 59 38 PM

(can compare to legacy and lsp4ij in https://github.com/flutter/dart-intellij-third-party/pull/357)


helin24:
@gemini-code-assist review again?

third_party/src/main/java/com/jetbrains/lang/dart/lsp/DartLspServerDescriptor.kt
```diff
@@ -0,0 +1,76 @@
...
+ * 1. Which files are supported (only `.dart` files).
+ * 2. The communication channel to use (a TCP Socket channel using the dynamically allocated port of [DartBridgeLspServerManager]).
+ *    `startProcess = false` tells the platform that the server is already running internally, so it only needs to connect.
+ * 3. The LSP feature customizations (e.g. enabling/disabling hover support dynamically based on settings).
+ */
+class DartLspServerDescriptor(project: Project) : ProjectWideLspServerDescriptor(project, "Dart (Bridge)") {
+
+    override fun isSupportedFile(file: VirtualFile): Boolean {
+        return file.extension == "dart"
+    }
```
alexander-doroshko, line 55: Dart Analysis Server understands not only '.dart' files. 

See `DartAnalysisServerService#isFileNameRespectedByAnalysisServer`. It might make sense to review the list of files by checking the Analysis Server source code. If the list of files is the same for Legacy mode and LSP mode then just reuse `DartAnalysisServerService#isFileNameRespectedByAnalysisServer`

third_party/src/main/java/com/jetbrains/lang/dart/lsp/DartBridgeLspServer.kt
```diff
@@ -0,0 +1,357 @@
...
+ * 2. **LSP-Only Handlers** (e.g., Go to Definition, Completion): These are NOT registered in legacy DAS.
+ *    If we forward them, DAS will return `MethodNotFound`. To support these, the bridge must manually
+ *    intercept the LSP request, translate it to a legacy DAS request (e.g., `analysis.getNavigation` for Definition),
+ *    and translate the legacy response back to LSP classes.
+ *
+ * ## Document Synchronization
+ * Standard LSP document sync notifications ([didOpen], [didChange], [didClose]) are ignored in this bridge.
+ * The legacy Dart plugin already handles robust file content synchronization via `analysis.updateContent`.
+ * Since DAS shares the same in-memory overlay filesystem between legacy and LSP handlers, the shared LSP handlers
+ * (like Hover) will automatically see the up-to-date content synced by the legacy plugin.
```
alexander-doroshko, line 62: I still think that calling `das.updateFileContent()` from `didOpen/Change/Close` might be required in some cases.

If `das.updateFileContent()` has already been called, then the second invocation will do nothing and return quickly - so no risk to call it once again.

In LSP, the order of requests and notifications is very important. Let's take, for example, auto-completion on typing. It's extremely important to sync file content before asking server for completions. Currently, `das.updateFilesContent()` is called by `DartServerCompletionContributor`. When you switch this feature to LSP, `DartServerCompletionContributor` won't work. `das.updateFilesContent()` may still happen from elsewhere but without guarantee that it will happen before asking server for completions. Calling `das.updateFilesContent()` from the bridge server's `didOpen/Change/Close` seems important.

`didSave` is a different story. I fthink you may keep it no-op unless Legacy server needs it

third_party/src/main/java/com/jetbrains/lang/dart/lsp/DartLspServerDescriptor.kt
helin24: Checked and these are the same; or more specifically, analysis server uses the same instance of `ContextManagerImpl`, which is for watching files.

third_party/src/main/java/com/jetbrains/lang/dart/lsp/DartBridgeLspServer.kt
helin24: True, I think I ran into this problem eventually with trying to implement codeAction with lsp4ij. I added an call to update before forwarding requests.

third_party/thirdPartySrc/platform-lsp/src/com/intellij/platform/dartlsp/impl/features/codeLens/LspCodeVisionProvider.kt
```diff
@@ -64,5 +64,7 @@ internal class LspCodeVisionProvider : CodeVisionProvider<Unit>, DumbAware {
   override val relativeOrderings: List<CodeVisionRelativeOrdering> = emptyList()
   override val id: String = LSP_CODE_VISION_PROVIDER_ID
   override val name: String = LspBundle.message("codeLens.LspCodeVisionProvider.name")
-  //override val singleEntryPerLine: Boolean = false
+  override val singleEntryPerLine: Boolean = false
```
helin24, line 67: This line and the one below are for 2025.3 verifier issues. The original methods of this class (which is part of the copied-over JetBrains LSP code) call platform code that triggers verifier problems from 2025.3 not having certain methods.

third_party/build.gradle.kts
```diff
@@ -101,7 +101,7 @@ intellijPlatform {
             VerifyPluginTask.FailureLevel.SCHEDULED_FOR_REMOVAL_API_USAGES,
             //            VerifyPluginTask.FailureLevel.EXPERIMENTAL_API_USAGES,
             //            VerifyPluginTask.FailureLevel.INTERNAL_API_USAGES,
-            VerifyPluginTask.FailureLevel.OVERRIDE_ONLY_API_USAGES,
+            //            VerifyPluginTask.FailureLevel.OVERRIDE_ONLY_API_USAGES,
```
helin24, line 104: It makes sense to turn this off, since we now have "platform code" in our plugin that use methods in the platform that are marked override-only. However, we should try to make sure that no new verifier baselines are added in the override-only category that are NOT related to LSP.

.agents/skills/patch-copied-lsp-sources/scripts/patch.py
```diff
@@ -233,6 +233,19 @@ def remove_ids(match):
     with open(properties_path, "w", encoding="utf-8") as f:
         f.write(properties_content)
 
+    # 9. Modify LspCodeVisionProvider.kt to explicitly override handleExtraAction
```
helin24, line 236: Note that this is within the skill to apply changes if we need to refresh the JB LSP code.

third_party/build.gradle.kts
pq: Why not add the new `OVERRIDE_ONLY_API_USAGES` violations to the baselines?

third_party/tool/baseline/261/verifier-baseline.txt
```diff
@@ -1,3 +1,5 @@
+* Deprecated class org.eclipse.lsp4j.MarkedString is referenced in com.intellij.platform.dartlsp.impl.features.documentation.TextRangeAndMarkupContent.Companion.fromHover$lambda$0(Either) : CharSequence
```
pq: This is adding *a lot* of new entries (> 400?!) Are they all necessary?

third_party/thirdPartySrc/platform-lsp/src/com/intellij/platform/dartlsp/impl/features/codeLens/LspCodeVisionProvider.kt
alexander-doroshko: Hmm,
There is no `singleEntryPerLine` property in the `CodeVisionProvider` class in the [branch 2025.3](https://github.com/JetBrains/intellij-community/blob/253/platform/lang-impl/src/com/intellij/codeInsight/codeVision/CodeVisionProvider.kt).

It is available only starting from [2026.1](https://github.com/JetBrains/intellij-community/blob/261/platform/lang-impl/src/com/intellij/codeInsight/codeVision/CodeVisionProvider.kt).

This code won't compile against 2025.3:
image

alexander-doroshko: Maybe, once compiled against 2026.1, the compiled class can work fine in 2025.3 as well, and overriding non-existing property in a compiled .class file doesn't cause issues in runtime... Need to check JVM behavior to be sure.

In any case, I wouldn't expect warnings from a 2025.3 verifier. What was the error/warning?

helin24: Sorry, this is confusing; hopefully I can explain adequately. We compile against 2026.1 only, but we check the resulting plugin with the 2025.3 verifier. If we had compiled against 2025.3 then I think neither of these would be issues.

The errors/warnings without making changes to this file were:
```
Invocation of unresolved method com.intellij.codeInsight.codeVision.CodeVisionProvider.getSingleEntryPerLine() : boolean [Compatibility problems]
```

and 
```
Method com.intellij.platform.dartlsp.impl.features.codeLens.LspCodeVisionProvider.getSingleEntryPerLine() : boolean contains an *invokespecial* instruction referencing an unresolved method com.intellij.codeInsight.codeVision.
  CodeVisionProvider.getSingleEntryPerLine() : boolean. This can lead to **NoSuchMethodError** exception at runtime.
```

third_party/build.gradle.kts
helin24: Adding to the baselines doesn't prevent verify from gradle verify failing when one of these is detected. When this is commented out in build file, verify will still write it to its output of verify issues and our CI will compare to the baseline files.

third_party/tool/baseline/261/verifier-baseline.txt
helin24: Hmm.. I think these are all necessary because of the copied code. I wonder if we could add a mechanism to ignore all verification issues coming from `com.intellij.platform.dartlsp`? Do you think something like that would be more preferable?

pq: Totally. That would be great.

helin24: I updated the presubmit and the baseline script to ignore `dartlsp` issues

third_party/src/main/java/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -78,6 +78,8 @@
 import org.jetbrains.annotations.NonNls;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
+import com.jetbrains.lang.dart.lsp.DartBridgeLspServerManager;
```
pq: Sort imports?

pq (approved):
Thanks!

third_party/src/main/java/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
helin24: Thanks, I fixed sorting imports across the changed files.

PR #467, opened 2026-06-11
Modify JetBrains LSP copied sources and create skill for repeating this

helin24:
These are follow up items from Alex's comments in https://github.com/flutter/dart-intellij-third-party/pull/442, mostly for keeping our copied from conflicting with anything in the IJ platform code:
- Rename connect timeout registry key to 'dart.lsp.server.connect.timeout' to prevent conflicts with IntelliJ's built-in key.
- Prefix LSP notification groups with 'Dart: ' and update the corresponding Kotlin constants to prevent conflicts.
- Remove redundant/optional 'id' attributes from extension tags in XML.
- Change the resource bundle from LspBundle to DartLspBundle and provide the bundle file.
- Change log file registry key default to false to prevent spamming idea.log.
- Rename LspIntention folders and XML configurations to DartLspIntention.
- Create patch-copied-lsp-sources skill under .agents/ to automate these patches. (This is for cases where we need to refresh these files later)


third_party/thirdPartySrc/platform-lsp/src/com/intellij/platform/dartlsp/api/lsWidget/LspServerWidgetItem.kt
```diff
@@ -17,13 +17,13 @@ import com.intellij.platform.lang.lsWidget.LanguageServicePopupSection.ForCurren
...
 import com.intellij.platform.lang.lsWidget.LanguageServiceWidgetItem
 import com.intellij.platform.lang.lsWidget.OpenSettingsAction
-import com.intellij.platform.lsp.dart.api.LspBundle
-import com.intellij.platform.lsp.dart.api.LspServer
-import com.intellij.platform.lsp.dart.api.LspServerManager
-import com.intellij.platform.lsp.dart.api.LspServerState.Initializing
-import com.intellij.platform.lsp.dart.api.LspServerState.Running
-import com.intellij.platform.lsp.dart.api.LspServerState.ShutdownNormally
-import com.intellij.platform.lsp.dart.api.LspServerState.ShutdownUnexpectedly
+import com.intellij.platform.dartlsp.api.LspBundle
```
pq, line 15: It looks like imports should all be sorted again after the rename?

helin24: This succeeded, and it also added newlines to files 🤨 I mean, I guess that's okay?

.agents/skills/patch-copied-lsp-sources/scripts/patch.py
```diff
@@ -0,0 +1,186 @@
...
+        else:
+            base_dir = os.getcwd()
+
+    if not os.path.isdir(os.path.join(base_dir, "third_party")):
+        print(f"Error: {base_dir} does not contain a 'third_party' folder.", file=sys.stderr)
+        sys.exit(1)
+
+    print(f"Applying JetBrains LSP patch with package rename to: {base_dir}")
+
+    # 1. Rename com/intellij/platform/lsp/dart to com/intellij/platform/dartlsp
```
pq: Do we need to commit this? Isn't the rename a one-time only thing?

.agents/skills/patch-copied-lsp-sources/scripts/patch.py
```diff
@@ -0,0 +1,186 @@
...
+
+    # 3. Modify dart-lsp-impl.xml
+    xml_path = os.path.join(base_dir, "third_party/thirdPartySrc/platform-lsp/resources/dart-lsp-impl.xml")
+    if os.path.exists(xml_path):
+        with open(xml_path, "r", encoding="utf-8") as f:
+            xml_content = f.read()
+
+        xml_content = xml_content.replace(
+            'key="lsp.server.connect.timeout"',
+            'key="dart.lsp.server.connect.timeout"'
```
pq, line 112: Should this be `dartlsp`?

.agents/skills/patch-copied-lsp-sources/scripts/patch.py
```diff
@@ -0,0 +1,186 @@
...
+            f.write(xml_content)
+
+    # 4. Modify Lsp4jServerConnector.kt
+    connector_path = os.path.join(base_dir, "third_party/thirdPartySrc/platform-lsp/src/com/intellij/platform/dartlsp/impl/connector/Lsp4jServerConnector.kt")
+    if os.path.exists(connector_path):
+        with open(connector_path, "r", encoding="utf-8") as f:
+            connector_content = f.read()
+        connector_content = connector_content.replace(
+            '"lsp.server.connect.timeout"',
+            '"dart.lsp.server.connect.timeout"'
```
pq, line 154: `lspdart`?

.agents/skills/patch-copied-lsp-sources/SKILL.md
```diff
@@ -0,0 +1,51 @@
...
+   ```bash
+   ./gradlew clean compileKotlin --no-build-cache
+   ```
+
+## Record of Adjustments Applied
+
+Here is the details of the changes made by the patch script:
+
+1. **Package Namespace Rename**:
+   - Rename package namespace `com.intellij.platform.lsp.dart` to `com.intellij.platform.dartlsp` across all `.kt`, `.java`, and `.xml` source files.
```
pq: Same question again. Does this need to be here if we only needed to do it this one time?

(Sorry if I'm being thick!)

.agents/skills/patch-copied-lsp-sources/scripts/patch.py
helin24: I don't know - I figured that between now and when 2026.2 is our lower bound may be 6 months to a year, and it seems likely there could be changes we need to incorporate from the jetbrains code. So I requested a skill to save this information of the minor adjustments we want to make on top of copying over the code.

In terms of the script, Gemini thinks having a python script is ultimately faster than having an agent rescan files based on text in a skill. 🤷

.agents/skills/patch-copied-lsp-sources/scripts/patch.py
helin24: This is for registry keys, which is different from the directory structure. The standalone `dart` shows that this key is associated with our plugin.

.agents/skills/patch-copied-lsp-sources/scripts/patch.py
helin24: same as above

.agents/skills/patch-copied-lsp-sources/SKILL.md
helin24: same shruggly answer as above 😅

.agents/skills/patch-copied-lsp-sources/scripts/patch.py
pq: Got it. But maybe it's good to stay consistent?

.agents/skills/patch-copied-lsp-sources/scripts/patch.py
pq: Ah! So this is what I'm missing. I thought `com/intellij/platform/lsp/dart` was the package you chose in your previous PR and that this rename was a one time only thing? Are you saying this rename will have to happen every time we update?

pq (approved):
I think I get it now!

.agents/skills/patch-copied-lsp-sources/scripts/patch.py
helin24: I think we don't want to use `dartlsp` here because we want the registry keys to be recognizable as part of our plugin which uses the namespace `dart`.

.agents/skills/patch-copied-lsp-sources/scripts/patch.py
helin24: Fixed now!

PR #469, opened 2026-06-11
Remove foojay resolver

helin24:
This was initially added for specifying a version of java, which was required for lsp4ij. Since we've removed lsp4ij, I think we should be able to remove foojay resolver as well. Instead, we will just depend on whatever version of java is on the build server and expect that it will be backward compatible.

### DIFF
--- a/third_party/build.gradle.kts
+++ b/third_party/build.gradle.kts
@@ -133,6 +133,9 @@ kotlin {
 }
 
 java {
+    toolchain {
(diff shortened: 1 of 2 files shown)


pq:
Maybe setting `javaVersion` to 21 in `gradle properties` is a problem and we should go for what we do in the flutter plugin?

```kotlin 
val javaVersion = providers.gradleProperty("javaVersion").get()
```

pq (approved):
🎉

PR #478, opened 2026-06-16
Report usage analytics for Dart run and debug sessions

helin24:
I've been looking into using JetBrains DAP, which would allow us to remove usages of the legacy analysis server for file mapping (this functionality does not exist in LSP analysis server because debug adapter in Dart SDK takes care of it). DAP seems promising, but I also want to check the usage of the current run/debug mechanisms. Then if we transition, we will be able to check whether we are getting equivalent usage, and the relative numbers of users may also inform how much time we spend on verifying DAP setups before releasing.


third_party/src/main/java/com/jetbrains/lang/dart/ide/runner/server/webdev/DartWebdevConfiguration.java
```diff
@@ -54,6 +55,7 @@ public DartWebdevConfiguration clone() {
   @Override
   public @Nullable RunProfileState getState(final @NotNull Executor executor,
                                             final @NotNull ExecutionEnvironment env) throws ExecutionException {
+    Analytics.recordRunOrDebugSession("web", executor, getProject());
```
pq: Could you extract `"web"` and the following into constants?

pq (approved):

Issue #481, opened 2026-06-21
Replace Weberknecht with JDK's Websocket in DartToolingDaemonService


JavadAsadi:
Relates to #208 

After the change, plugin is ready to have Dart Tooling Daemon with auto ping intervals.
The decision to enable the ping interval on the plugin side with command line or on the sdk side will be made outside the scope of this issue.

PR #482, opened 2026-06-21
Replace Weberknecht with JDK's Websocket in DartToolingDaemonService.kt

JavadAsadi:
Solves #481

This PR enables the plugin to handle ping/pong requests. Whether the `commandLine.addParameter("--ping-interval=15")` is needed to be added in the code or just add the ping-interval on the sdk side will be decided later.


helin24 (approved):
LGTM!

pq (approved):
Awesome!

PR #488, opened 2026-06-23
[2025.2 branch] Restore expected diagnostic in ErrorsAfterEOF.dart

helin24:
I think this was accidentally deleted and causing integration tests to fail.


helin24:
I think this still has failures due to timeouts, so I will close this and move the change to https://github.com/flutter/dart-intellij-third-party/pull/487.

PR #492, opened 2026-06-26
Support LSP hover for external library files (pub-cache, SDK)

helin24:
The initial change using LSP makes hover unavailable to library files, e.g. if we open a Flutter framework file that an app depends on. Since legacy LSP provides this, I think we should enable it for these other files if possible. (To check this change, open up a Flutter SDK file and make sure hover works in that file as well.)

The current change is one in the copied LSP code. I'm wondering if there's a possibility of making this available as an option in the public-facing API so that we can have this in the future while using the open source version? @alexander-doroshko

### DIFF
--- a/.agents/skills/patch-copied-lsp-sources/SKILL.md
+++ b/.agents/skills/patch-copied-lsp-sources/SKILL.md
@@ -55,3 +55,6 @@ Here is the details of the changes made by the patch script:
 
 7. **Intention Directory Names**:
    - Rename the intention descriptions folder `LspIntention` to `DartLspIntention` and update `descriptionDirectoryName` to `DartLspIntention` in `dart-lsp-impl.xml` to prevent collision.
+
+8. **Support External Library Files**:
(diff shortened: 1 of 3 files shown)


pq:
I'd be a little hesitant to land this as it's functionality we'll lose when we stop copying and use the libraries directly.

Can you add this fix by subclassing or elsewhere in non-platform code? (I haven't looked TBH.)

helin24:
I didn't find a way to fix this without changing the platform code, but I'll reach out to Alex to see if this is something we can introduce to platform code.

alexander-doroshko:
Thanks for drawing my attention to this change. Indeed, this seems to be a limitation of the IntelliJ's LSP Client. I agree that we should remove this limitation, either in the same way as in this PR, or in some other way. In the meantime, I think it's reasonable to proceed with this change in the Dart plugin.

pq (approved):
🚀

PR #493, opened 2026-06-26
Add custom request for diagnostics port

helin24:
This adds a custom request to retrieve the diagnostic port, used here:
image

This also shows an example of checking the SDK version to make sure that the analysis server supports the LSP-over-legacy request. (As we are transitioning more things to LSP, ask me for a doc of when SDK supports LSP-over-legacy requests - most have been available for >2 years, but there are some exceptions that will require a version check).

(This was previously implemented with lsp4ij back in https://github.com/flutter/dart-intellij-third-party/pull/391/changes)

### DIFF
--- a/third_party/src/main/java/com/jetbrains/lang/dart/lsp/LspMethod.kt
+++ b/third_party/src/main/java/com/jetbrains/lang/dart/lsp/LspMethod.kt
@@ -12,7 +12,8 @@ enum class LspMethod(
 ) {
     INITIALIZE("initialize"),
     SHUTDOWN("shutdown"),
-    HOVER("textDocument/hover", isExperimental = true, presentableName = "hover");
+    HOVER("textDocument/hover", isExperimental = true, presentableName = "hover"),
+    DIAGNOSTIC_SERVER("dart/diagnosticServer", isExperimental = true, presentableName = "diagnostic server");
 
     companion object {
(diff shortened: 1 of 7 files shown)


third_party/src/main/java/com/jetbrains/lang/dart/lsp/LspMethod.kt
```diff
@@ -12,7 +12,8 @@ enum class LspMethod(
 ) {
     INITIALIZE("initialize"),
     SHUTDOWN("shutdown"),
-    HOVER("textDocument/hover", isExperimental = true, presentableName = "hover");
+    HOVER("textDocument/hover", isExperimental = true, presentableName = "hover"),
```
pq, line 14: Nit: I'd be inclined to alphabetically sort these unless there's some kind of semantic grouping going on.

pq (approved):

third_party/src/main/java/com/jetbrains/lang/dart/lsp/LspMethod.kt
helin24: Hmm.. I was going to have these in order of being added since we will be changing isExperimental to false. But I think it's fine to have them ordered alphabetically too and just check on what release they went out in.

PR #516, opened 2026-07-08
Replace Weberknecht with JDK's Websocket in VmServiceBase and WebSock…

JavadAsadi:
## Summary

This is the last service using the Weberknecht. After this pull request there will be no usage of Weberknecht in Dart plugin and we can safely remove the dependecy.

I had some manual tests to make sure VMService is working as expected with the new Websocket.

The only change I added to logic was creating the Websocket object with no try catch expression.

## Manual test

Use the plugin to debug a dart script. Intelij plugin uses a command like this:
```
dart
    --enable-asserts
    --pause_isolates_on_start
    --enable-vm-service:50788
    --no-serve-devtools
   /path/to/your/script
```
this will automatically use the VmService and if the application runs succesfully it means VMService connection has been succesful.

In order to check the messages put some breakpoints in the Websocket class. Check the messages to see if everything is working fine.

More edge cases are tested by a script in notebook


helin24 (approved):
Yay!

helin24:
@pq just CCing you for visibility. I'll merge as long as tests pass

Issue #208, opened 2026-01-13
[dtd] Implement Websocket Ping/Pong to Play Nice w/ Norton


pq:
DTD can be run with a setting that enables regular pings to clients to keep the DTD websocket connection alive in the face of anti-virus tools such as Norton that unceremoniously closr what it interprets as idle socket (https://github.com/Dart-Code/Dart-Code/issues/5794). The rub is that our current websocket implementation does not support automatic pongs in response to pings and does not surface low-enough level APIs to allow us to do it ourselves (https://github.com/flutter/dart-intellij-third-party/issues/205).

Until we address this, Norton users will face the same kinds of DTD interruptions as documented in https://github.com/Dart-Code/Dart-Code/issues/5794

Issue #205, opened 2026-01-12
[dtd] DTD Process Socket closing unexpectedly


pq:
The Dart Plugin initializes a `DartToolingDaemonService` which is meant to be long-running and the way that devtools and analytics interact with a DTD instance. In testing last week, I noticed that the service's websocket is reliably closing soon after initialization and it's not obvious why.

To repro, check out this branch which disables analytics (to keep things simple) and adds some logging: https://github.com/flutter/dart-intellij-third-party/pull/206

Then run the plugin and watch the console log.  What I'm seeing reliably is something like this:

```
2026-01-12 09:01:22,026 [  11549]   WARN - #c.j.l.d.i.t.DartToolingDaemonService - --> {"jsonrpc":"2.0","method":"registerService","id":1,"params":{"service":"Editor","method":"getActiveLocation","capabilities":{}}}
2026-01-12 09:01:22,027 [  11550]   WARN - #c.j.l.d.i.t.DartToolingDaemonService - --> {"jsonrpc":"2.0","method":"FileSystem.setIDEWorkspaceRoots","id":2,"params":{"roots":["file:///Users/pquitslund/src/tmp/dart_proj"],"secret":"0BdD3UM5vG4="}}
2026-01-12 09:01:22,027 [  11550]   WARN - #c.j.l.d.i.t.DartToolingDaemonService - --> {"jsonrpc":"2.0","method":"registerService","id":3,"params":{"service":"Editor","method":"navigateToCode","capabilities":{}}}
2026-01-12 09:01:22,032 [  11555]   WARN - #c.j.l.d.i.t.DartToolingDaemonService - <-- {"jsonrpc":"2.0","result":{"type":"Success"},"id":1}
2026-01-12 09:01:22,032 [  11555]   WARN - #c.j.l.d.i.t.DartToolingDaemonService - <-- {"jsonrpc":"2.0","result":{"type":"Success"},"id":2}
2026-01-12 09:01:22,032 [  11555]   WARN - #c.j.l.d.i.t.DartToolingDaemonService - <-- {"jsonrpc":"2.0","result":{"type":"Success"},"id":3}
2026-01-12 09:01:37,030 [  26553]   WARN - #c.j.l.d.i.t.DartToolingDaemonService - <-- 
[...]

PR #525, opened 2026-07-16
Enable ping interval for Dart tooling Daemon

JavadAsadi:
# Summary

Relates to #208 [Norton](https://github.com/Dart-Code/Dart-Code/issues/5794) #205 

- Enabling ping intervals for Dart Tooling Daemon keeps the connection reliable while the daemon is idle.

- It prevents proxies, VPNs, antivirus software, and network devices from dropping an apparently inactive connection.
- If DTD doesn’t receive a pong within the next interval, it considers the connection dead and closes it, allowing reconnection instead of retaining a stale socket. This follows Dart’s
    WebSocket ping behavior (https://api.dart.dev/dart-io/WebSocket/pingInterval.html).

### DIFF
--- a/third_party/src/main/java/com/jetbrains/lang/dart/dtd/DTDProcess.kt
+++ b/third_party/src/main/java/com/jetbrains/lang/dart/dtd/DTDProcess.kt
@@ -130,6 +130,7 @@ class DTDProcess {
     commandLine.charset = StandardCharsets.UTF_8
     commandLine.addParameter("tooling-daemon")
     commandLine.addParameter("--machine")
+    commandLine.addParameter("--ping-interval=15")
     Analytics.updateEnvironment(commandLine)
 
     osProcessHandler = object : KillableProcessHandler(commandLine) {
@@ -230,7 +231,7 @@ class DTDProcess {
       if (processRunning) return
 
       // The first line of text is the command issued, which can be ignored.
-      val text = event.text.trim().takeUnless { it.endsWith(" tooling-daemon --machine") }
+      val text = event.text.trim().takeUnless { it.endsWith(" tooling-daemon --machine --ping-interval=15") }
         ?: return
 
       var uri: String? = null
@@ -255,4 +256,3 @@ interface DTDProcessListener {
   fun onWebSocketOpen() {}
   fun onProcessStarted(uri: String?) {}
 }
-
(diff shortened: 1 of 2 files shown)


third_party/src/main/java/com/jetbrains/lang/dart/dtd/DTDProcess.kt
```diff
@@ -230,7 +231,7 @@ class DTDProcess {
       if (processRunning) return
 
       // The first line of text is the command issued, which can be ignored.
-      val text = event.text.trim().takeUnless { it.endsWith(" tooling-daemon --machine") }
+      val text = event.text.trim().takeUnless { it.endsWith(" tooling-daemon --machine --ping-interval=15") }
         ?: return
```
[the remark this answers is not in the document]
JavadAsadi: This should be fine for now. The idea is to revert it when ping interval is 15 by default on the sdk side.

third_party/src/main/java/com/jetbrains/lang/dart/ide/toolingDaemon/DartToolingDaemonService.kt
```diff
@@ -300,7 +301,7 @@ class DartToolingDaemonService private constructor(val project: Project, cs: Cor
       if (serviceRunning) return
 
       // The first line of text is the command issued, which can be ignored.
-      val text = event.text.trim().takeUnless { it.endsWith(" tooling-daemon --machine") }
+      val text = event.text.trim().takeUnless { it.endsWith(" tooling-daemon --machine --ping-interval=15") }
                  ?: return
```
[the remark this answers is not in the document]
JavadAsadi: This should be fine for now. The idea is to revert it when ping interval is 15 by default on the sdk side.

helin24 (approved):

helin24:
I'll go ahead and merge. CC @pq if you have any follow up comments.

If it's easy, you could also consider making a follow up PR that pulls out the command line arguments into constants, so that we don't have to repeat the argument strings when searching for the command printout to skip that line when parsing output.

third_party/src/main/java/com/jetbrains/lang/dart/dtd/DTDProcess.kt
```diff
@@ -130,6 +130,7 @@ class DTDProcess {
     commandLine.charset = StandardCharsets.UTF_8
     commandLine.addParameter("tooling-daemon")
     commandLine.addParameter("--machine")
+    commandLine.addParameter("--ping-interval=15")
```
pq, line 133: As per @helin24's comment, it would be great to pull this value out and add it as a parameter.

helin24: done in https://github.com/flutter/dart-intellij-third-party/pull/527

helin24: for future reference @JavadAsadi in the comments of the follow up PR, you can refer to the original PR to note that they're connected (and then we will be able to see in the original PR that there was a follow up)

PR #526, opened 2026-07-16
Implement LSP code actions

helin24:
These are the things that code actions handle:
- Quick fixes:
  - Import missing library
  - Create missing declaration (e.g. a class that doesn’t exist)
  - Add required parameter (e.g. you are trying to call a method but don’t remember params)
  - Remove dead/unused code
- Quick assists and structural refactorings
  - Flutter widget wrap and remove
  - Convert getter to method, vice versa
  - Convert positional to named parameters
- Source / batch actions
  - Organize imports (note - this is different from optimize imports)
  - Sort members (e.g. methods in a class that are not alphabetized)
- Interactive code refactorings
  - Extract method
  - Extract local variable

There are some significant UI differences from legacy, and we should decide if we are okay going ahead with these:
- Source / batch actions are newly in the option+enter menu. Before they needed shortcuts or Code menu
- Extract method in the legacy implementation offered options of whether you want to replace all occurrences and whether you want to make the method a getter.
- Extract local variable would ask you whether you wanted to replace all occurrences
- (Similarly, inline would ask whether you wanted to delete the variable/method)
- Cmd + Opt + M will no longer trigger refactor

In general, `edit.getRefactoring` usage has declined pretty significantly (for more details, ask me for this data), so I am not so worried about removing this.


third_party/src/main/java/com/jetbrains/lang/dart/lsp/DartBridgeLspServer.kt
```diff
@@ -254,6 +284,68 @@ class DartBridgeLspServer(private val project: Project) : DartLanguageServer, Te
...
+        val rawTree = GSON.toJsonTree(params)
+        logger.info("Client executeCommand called: command=${params.command}, rawParams=$rawTree")
+
+        val forwardedParams = normalizeExecuteCommandParams(params)
+
+        logger.info("Forwarding executeCommand to DAS: command=${params.command}, unpackedParams=${GSON.toJson(forwardedParams)}")
+        return forwardRequest("workspace/executeCommand", forwardedParams, Any::class.java)
+    }
+
+    private fun normalizeExecuteCommandParams(params: ExecuteCommandParams): ExecuteCommandParams {
```
helin24, line 372: @alexander-doroshko Gemini generated this for me, and from follow up conversation, I gather this is because we are using this custom bridge instead of letting the LSP client communicate with DAS over stdin/out. But it also seems there must be some code within LSP client that would do this kind of transformation to JSON; is that possible to call from our code, or possible to expose somehow, so that we don't need to have this custom code in the bridge server?

third_party/src/main/java/com/jetbrains/lang/dart/lsp/DartLspJsonConverter.kt
```diff
@@ -0,0 +1,166 @@
...
+import org.eclipse.lsp4j.jsonrpc.json.JsonRpcMethod
+import org.eclipse.lsp4j.jsonrpc.json.MessageJsonHandler
+import org.eclipse.lsp4j.jsonrpc.messages.Either
+import org.eclipse.lsp4j.services.LanguageClient
+import org.eclipse.lsp4j.services.LanguageServer
+import org.eclipse.lsp4j.services.TextDocumentService
+import org.eclipse.lsp4j.services.WorkspaceService
+import java.lang.reflect.ParameterizedType
+
+/**
```
helin24: @alexander-doroshko same question as above for this JSON transformation class.

third_party/src/main/java/com/jetbrains/lang/dart/lsp/DartBridgeLspServer.kt
```diff
@@ -178,17 +194,20 @@ class DartBridgeLspServer(private val project: Project) : DartLanguageServer, Te
...
                 val paramsObj = msgObj.get("params")
                 val params = GSON.fromJson(paramsObj, PublishDiagnosticsParams::class.java)
                 client.publishDiagnostics(params)
+            } else if (method == "workspace/applyEdit") {
+                val paramsObj = msgObj.get("params")
+                val params = GSON.fromJson(paramsObj, ApplyWorkspaceEditParams::class.java)
+                client.applyEdit(params)
             } else {
-                logger.info("Ignored notification from DAS: $method")
+                logger.info("Ignored notification/request from DAS: $method")
```
alexander-doroshko: Maybe `logger.debug()` or `logger.trace()` here and everywhere?

'Info' level is written to log files out-of-the-box, so all users will get megabytes of unneeded spam in logs.

'Debug' and 'Trace' levels are turned off by default, you can enable then only for yourself via Help -> Diagnostic Tools -> Debug Log Settings

third_party/src/main/java/com/jetbrains/lang/dart/lsp/DartBridgeLspServer.kt
```diff
@@ -179,17 +195,20 @@ class DartBridgeLspServer(private val project: Project) : LanguageServer, TextDo
...
-            // Currently, only 'textDocument/publishDiagnostics' is supported and forwarded.
+            // Parse and forward notifications/requests to the LSP client proxy using lsp4j.
             if (method == "textDocument/publishDiagnostics") {
                 val paramsObj = msgObj.get("params")
                 val params = GSON.fromJson(paramsObj, PublishDiagnosticsParams::class.java)
                 client.publishDiagnostics(params)
+            } else if (method == "workspace/applyEdit") {
+                val paramsObj = msgObj.get("params")
+                val params = GSON.fromJson(paramsObj, ApplyWorkspaceEditParams::class.java)
+                client.applyEdit(params)
```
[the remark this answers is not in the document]
alexander-doroshko: > but it completely ignores the returned CompletableFuture and never sends the LSP response back to the Dart Analysis Server (DAS)

Valid comment. This function is called `forwardNotificationToClient`. Perhaps there should be something like `forwardRequestToClient` as well, which would care about handling the client's response.

By the way, functions `handleDasResponse` and `setupDasResponseListener` have misleading names as they are currently handling not only DAS responses but also server-initiated notifications and requests.

third_party/src/main/java/com/jetbrains/lang/dart/lsp/DartBridgeLspServer.kt
alexander-doroshko: There is no such low-level JSON handling in the LSP Client code. It might be hidden somewhere in the LSP4J library, but we never had to care about it. 

I was a bit surprised to see the code that Gemini generated in `normalizeExecuteCommandParams` and in `DartLspJsonConverter`. I didn't expect the need of such low-level JSON handling.

I agree with Gemini as a reviewer (a comment above) that the code generated by Gemini as a developer looks fragile.

I'd expect the LSP4J library to contain all the low-level stuff we may need, but I haven't tried to simplify or remove the `normalizeExecuteCommandParams` /  `DartLspJsonConverter` code yet.

third_party/src/main/java/com/jetbrains/lang/dart/lsp/DartLspJsonConverter.kt
```diff
@@ -0,0 +1,166 @@
...
+                            return Either.forRight(rightVal)
+                        }
+                        if (CodeAction::class.java.isAssignableFrom(leftRaw) && Command::class.java.isAssignableFrom(rightRaw)) {
+                            if (obj.has("command") && !obj.has("kind") && !obj.has("edit") && !obj.has("diagnostics")) {
+                                val rightVal = rightAdapter.fromJsonTree(element)
+                                return Either.forRight(rightVal)
+                            }
+                            val leftVal = leftAdapter.fromJsonTree(element)
+                            return Either.forLeft(leftVal)
+                        }
```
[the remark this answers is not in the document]
helin24: This file was removed after I asked agent to remove extra code associated with having top-level command objects returned (since we are enabling code action literal)

third_party/src/main/java/com/jetbrains/lang/dart/lsp/DartLspJsonConverter.kt
helin24: This code has been removed, see earlier comment

third_party/src/main/java/com/jetbrains/lang/dart/lsp/DartBridgeLspServer.kt
helin24: Thanks, changing more of these to debug logs.

third_party/src/main/java/com/jetbrains/lang/dart/lsp/DartBridgeLspServer.kt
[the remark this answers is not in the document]
helin24: Fixed missing response and renamed these methods

third_party/src/main/java/com/jetbrains/lang/dart/lsp/DartBridgeLspServer.kt
helin24: I think this was all removed as a result of simplifying the code since the older top-level commands are no longer returned.

third_party/src/main/java/com/jetbrains/lang/dart/lsp/DartBridgeLspServer.kt
```diff
@@ -146,75 +166,102 @@ class DartBridgeLspServer(private val project: Project) : DartLanguageServer, Te
...
+                }
+                val legacyRequest = JsonObject().apply {
+                    addProperty("id", legacyId)
+                    addProperty("method", "lsp.handle")
+                    add("params", JsonObject().apply {
+                        add("lspMessage", lspResponse)
+                    })
+                }
+                try {
+                    das.sendRequest(legacyId, legacyRequest)
```
alexander-doroshko, line 255: Shouldn't it be `das.sendResponse()` that would it its turn call `RemoteAnalysisServerImpl#sendResponseToServer`? Technically, it's a response from the client to the server, but I didn't check precisely what the Legacy-over-LSP protocol requires here

third_party/src/main/java/com/jetbrains/lang/dart/lsp/DartBridgeLspServer.kt
```diff
@@ -279,6 +328,19 @@ class DartBridgeLspServer(private val project: Project) : DartLanguageServer, Te
...
+    override fun codeAction(params: CodeActionParams): CompletableFuture<List<Either<Command, CodeAction>>> {
+        val responseType = object : TypeToken<List<CodeAction>>() {}.type
+        return forwardRequest<List<CodeAction>>("textDocument/codeAction", params, responseType).thenApply { actions ->
+            actions?.map { Either.forRight<Command, CodeAction>(it) } ?: emptyList()
+        }
+    }
+
+    override fun resolveCodeAction(unresolved: CodeAction): CompletableFuture<CodeAction> {
+        return CompletableFuture.completedFuture(unresolved)
+    }
```
alexander-doroshko, line 342: Am I right that this function is never called because the Dart LSP server doesn't advertise CodeAction resolve capability?

helin24: Yes, that seems to be correct. (As a note to myself / future readers, resolve is a server capability, indicating that the server can send actions that are not fully formed edits but wait for more user intention before sending a full edit)

third_party/src/main/java/com/jetbrains/lang/dart/lsp/DartBridgeLspServer.kt
helin24: Thanks, it seems this wasn't sending the right response.

third_party/src/main/java/com/jetbrains/lang/dart/ide/annotator/DartAnnotator.java
```diff
@@ -194,10 +195,12 @@ public void annotate(final @NotNull PsiElement element, final @NotNull Annotatio
       }
     }
 
-    processDartRegionsInRange(notYetAppliedErrors, element.getTextRange(), err -> {
-      VirtualFile vFile = element.getContainingFile().getVirtualFile();
-      createAnnotation(holder, err, new DartQuickFixSet(element.getManager(), vFile, err.getOffset(), err.getCode()));
-    });
+    if (!DartAnalysisServerService.isLspPublishDiagnosticsEnabled(element.getProject())) {
```
helin24, line 198: `DartAnnotator` provides squiggles for errors and warnings. As part of code actions, we are going to get quick fixes associated with the errors through LSP, so continuing to have these in legacy would cause a list of duplicated quick fixes.

Code actions is enabled in an earlier version of SDK than publish diagnostics, but gating should be on publish diagnostics to coordinate the use of LSP for quick fixes.

pq (approved):

Issue #159, opened 2025-12-17
Add Inlay Hints to Dart code: variable type, parameter name, and more.


pq:
See corresponding issue in youtrack: https://youtrack.jetbrains.com/issue/IDEA-283678/Add-Inlay-Hints-to-Dart-code-variable-type-parameter-name-and-more.

helin24:
We should consider this closed with #617

Issue #400, opened 2026-05-13
[lsp] Change closing labels to LSP


helin24:
Legacy implementation is in `DartInlayHintsProvider`. However, may need investigation on whether inlay hints accomplishes the same things as closing labels.

helin24:
Note from Danny: We don't use inlay hints for closing labels currently (I wanted to, but the LSP spec is not flexible enough). Instead, the server publishes a custom notification ("publishClosingLabels") that Dart-Code listens for and updates them.

For now, that means we'd need some equivalent work in the IJ plugin to accept the same.

(note: we do use inlay hints for other things though - parameter names/types)

ralph-bergmann:
Picking this up from the note above — I'd like to do the plugin side, but the LSP-over-Legacy transport needs a decision first (cc @DanTup).

Current state, as far as I can see:
- Legacy: the plugin subscribes `CLOSING_LABELS` for visible files (`analysis_setSubscriptions`) and `DartServerData.computedClosingLabels` feeds the (renamed) `DartClosingLabelsInlayHintsProvider`; the *Editor | Inlay Hints | Closing labels* toggle only switches rendering.
- LSP: `dart/textDocument/publishClosingLabels` is sent only by `LspAnalysisServer`, for priority files, and only when the client passed the initialization option `closingLabels` (`lsp_analysis_server.dart`, `shouldSendClosingLabelsFor`). The legacy server never sends it, and there is no `initializationOptions` over LoL.
- Precedent for such a switch: dart-lang/sdk#64021 → `d42063aac44` sends `textDocument/publishDiagnostics` to LoL clients that advertise the `publishDiagnostics` client capability. Closing labels have no standard capability to piggy-back on.

The plugin side, once the server sends it, is small: the bridge forwards the notification into `DartServerData.computedClosingLabels` (LSP ranges → offsets); provider and `DartClosingLabelManager` stay as they are.

Questions:
1. Do you want closing labels migrated now (still on LoL), or only when the plugin talks pure LSP?
[...]

PR #551, opened 2026-07-30
Rename DartInlayHintsProvider to DartClosingLabelsInlayHintsProvider

ralph-bergmann:
The class only renders closing labels; with LSP inlay hints planned (#159) the old name becomes misleading. `providerId` ("dart.closing.labels") and all registration attributes are unchanged, so persisted user settings are unaffected. No functional change.

The "Flutter widget closing label" still works. Nothing more to test.

Disclaimer: This was mainly developed by Claude Fable 5 with the help of [superpowers](https://github.com/obra/superpowers)

---

<details>

  `dart format`.

</details>

### DIFF
--- a/third_party/CHANGELOG.md
+++ b/third_party/CHANGELOG.md
@@ -6,7 +6,6 @@
 
 ### Changed
 - Go to Definition implemented with JetBrains LSP (#539)
-- Rename `DartInlayHintsProvider` to `DartClosingLabelsInlayHintsProvider` to reflect that it only renders closing labels; provider ID and user settings are unchanged (#551)
 
 ### Removed
 
--- a/third_party/CHANGELOG.md
+++ b/third_party/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 ### Changed
 - Go to Definition implemented with JetBrains LSP (#539)
-- Renamed `DartInlayHintsProvider` to `DartClosingLabelsInlayHintsProvider` — it only renders closing labels; provider ID and user settings are unchanged (#551)
+- Rename `DartInlayHintsProvider` to `DartClosingLabelsInlayHintsProvider` to reflect that it only renders closing labels; provider ID and user settings are unchanged (#551)
 
 ### Removed
 
--- a/third_party/CHANGELOG.md
+++ b/third_party/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 ### Changed
 - Go to Definition implemented with JetBrains LSP (#539)
+- Renamed `DartInlayHintsProvider` to `DartClosingLabelsInlayHintsProvider` — it only renders closing labels; provider ID and user settings are unchanged (#551)
 
 ### Removed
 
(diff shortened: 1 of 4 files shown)


third_party/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 ### Changed
 - Go to Definition implemented with JetBrains LSP (#539)
+- Renamed `DartInlayHintsProvider` to `DartClosingLabelsInlayHintsProvider` — it only renders closing labels; provider ID and user settings are unchanged (#PR)
```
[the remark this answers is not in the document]
ralph-bergmann: fixed

third_party/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 ### Changed
 - Go to Definition implemented with JetBrains LSP (#539)
+- Rename `DartInlayHintsProvider` to `DartClosingLabelsInlayHintsProvider` to reflect that it only renders closing labels; provider ID and user settings are unchanged (#551)
```
helin24: Let's remove this since it isn't a user-facing change.

helin24 (changes requested):
I think this is reasonable, though I expect we will aim to remove this class once we have inlay hints. Based on preliminary research, inlay hints is at the beginnings of classes, but potentially can be put at the ends of classes with certain options?

ralph-bergmann:
You're right that the protocol would allow it — an `InlayHint` is just `{position, label}`, so a server could emit closing labels at end-of-line positions. But the Dart server currently doesn't: `textDocument/inlayHint` (`computer_inlay_hint.dart`) only computes type/parameter hints, and there's no option to include closing labels. They're served via the separate custom notification `dart/textDocument/publishClosingLabels` (opt-in through `initializationOptions.closingLabels`).

Interestingly, folding closing labels into inlay hints was considered on the SDK side when LSP inlay hints were specced (Dart-Code/Dart-Code#3628, 2021), but it didn't happen: VS Code still renders closing labels from the custom notification via decorations, with a separate toggle, color, and `//` prefix. I couldn't find any SDK issue planning to merge them.

So my expectation: once LSP inlay hints land (#159), both coexist — the generic LSP provider renders type/parameter hints while this class keeps rendering closing labels. When closing labels migrate to LSP (#400), this class switches its data source to the custom notification but is still needed as the renderer (the platform's generic inlay-hint support only serves `textDocument/inlayHint`). Removing it entirely would require an SDK change to include closing labels in `inlayHint` responses — happy to file that upstream if you think it's the right direction, but it would couple two features that VS Code deliberately keeps separately configurable. This answers the open investigation question in #400, by the way.

third_party/CHANGELOG.md
ralph-bergmann: Removed.

helin24 (approved):
Thanks for the clarification! I'll leave this for @pq to review and merge.

pq (approved):

DanTup:
> Removing it entirely would require an SDK change to include closing labels in `inlayHint` responses

FWIW I wanted to do this, but currently VS Code doesn't support custom kinds of inlay hints, so the styling of Closing Labels would become the same as other Inlay Hints and the customisation would be lost.

The VS Code issue I filed about this is at https://github.com/microsoft/vscode/issues/151920. If VS Code and LSP gain support for custom kinds (that we can customise the style of in the client), and IntelliJ could support the same, we should revisit it.

PR #573, opened 2026-08-05
Fix `.new` dot shorthand arg completion

### LINKED ISSUE #543: Autocomplete suggestions work inside .named() but not .new() constructor shorthand invocations
Transferred from: https://github.com/dart-lang/sdk/issues/63899

---
IntelliJ 2026.2.0.1
Dart plugin version: 507.0.0
Dart SDK version: 3.12.2
Flutter SDK version: 3.44.8

```dart

class Foo {
  Foo.named({String? name});
  Foo({String? name});
}

void a(Foo f) {}

void b() {
  // ✓ suggests name:
  a(.named(^));

  // ✓ suggests name:
  a(.named(na^));

  // × No suggestions
  a(.new(^));

  // × No suggestions
  a(.new(na^));
}

```

It actually works as expected in current DartPad with the same Dart/Flutter versions, so it seems specific to IntelliJ (haven't tried VS Code).

jakemac53:
Fixes #543                                                                                                                                                                                                  
              
 ## Root Cause                                                                                                                                                                                             
                                                                                                                                                                                                              
  In Dart.bnf, shorthandNewExpressionPrefix (which parses .new) was defined as a private Grammar-Kit rule. Because private rules do not produce composite AST nodes, the . and 'new' tokens were inlined      
  directly into DartNewExpression as leaf tokens.                                                                                                                                                             
                                                                                                                                                                                                              
  Without a child DartReference AST node to represent the .new constructor designator, DartReferenceImpl#getRangeInElement() fell back to returning the text range of the entire expression—including the     
  trailing argument list (...). When code completion was invoked inside .new(...), IntelliJ assumed the cursor was positioned mid-reference, computed an incorrect prefix (e.g., .new(), and discarded valid  
  suggestions returned by the Analysis Server.                                                                                                                                                                
                                                                                                                                                                                                              
 ## Solution                                                                                                                                                                                               
                                                                                                                                                                                                              
  This PR solves the problem purely at the grammar level in accordance with standard IntelliJ language plugin design:                                                                                         
                                                                                                                                                                                                              
  • Grammar Update (Dart.bnf): Made shorthandNewExpressionPrefix a public rule implementing DartReference. Grammar-Kit now generates an explicit DartShorthandNewExpressionPrefix AST node for .new, allowing 
  getRangeInElement() to target the constructor reference cleanly without enclosing the argument list.                                                                                                        
  • Parser Regeneration: Regenerated DartParser and visitor interfaces in gen/ to incorporate DartShorthandNewExpressionPrefix.                                                                               
  • IDE Handlers: Updated DartParameterInfoHandler and DartDeclarationRangeHandler to recognize DartShorthandNewExpressionPrefix alongside standard reference expressions, ensuring parameter tooltips and    
  declaration highlighting continue to work natively with shorthand constructors.                                                                                                                             
  • Revert Java Workaround: Reverted earlier defensive range exclusion workarounds in DartReferenceImpl, preserving clean baseline reference calculation.                                                     
                                                                                                                                                                                                              
  ## Testing & Verification                                                                                                                                                                                 
                                                                                                                                                                                                              
  Added regression test cases to DartDotShorthandCompletionTest.java:                                                                                                                                         
                                                                                                                                                                                                              
  • testDotNewArgCompletion(): Verifies argument autocompletion inside .new(<caret>)                                                                                                                          
  • testDotNewPartialArgCompletion(): Verifies argument autocompletion with partial input (.new(na<caret>))                                                                                                   
  • testDotNewWithWhitespaceAndCommentsArgCompletion(): Verifies robustness against intervening formatting (.new /* comment */ (<caret>))                                                                     
  • testDotNamedArgCompletion() & testDotNamedPartialArgCompletion(): Verifies parameter completion inside named shorthand constructor calls (.named(<caret>))  
  
 Updates to existing tests                                                                                                                                                                         
                                                                                                                              
  • DartParsingTest: The .txt files in testData/parsing/ have been properly updated to expect the new DartShorthandNewExpressionPrefix AST structure.

### DIFF
--- a/third_party/src/test/java/com/jetbrains/dart/analysisServer/DartDotShorthandCompletionTest.java
+++ b/third_party/src/test/java/com/jetbrains/dart/analysisServer/DartDotShorthandCompletionTest.java
@@ -64,22 +64,44 @@ public class DartDotShorthandCompletionTest extends DartServerCompletionTest {
              }""");
   }
 
+  public void testDotNamedPartialArgCompletion() {
+    doTest("name: ",
+           """
+             class A {
+               A.named({String? name});
+             }
+             void f(A a) {}
+             main() {
+               f(.named(na<caret>));
+             }""");
+  }
+
+  public void testDotNewPartialArgCompletion() {
+    doTest("name: ",
+           """
+             class A {
+               A({String? name});
+             }
+             void f(A a) {}
+             main() {
+               f(.new(na<caret>));
+             }""");
+  }
+
   private void doTest(String lookupToSelect, String text) {
     myFixture.configureByText("foo.dart", text);
     myFixture.doHighlighting();
     myFixture.completeBasic();
-    myFixture.type(lookupToSelect); 
-    // We expect the completion to work, so we check if the item is in the list
-    // The base class methods might be useful but we need custom setup for text verification or just basic completion check
-    // Actually, checking if "new" or "named" is present is enough for now.
-    // But verify behavior: if we type "new", it should select "new"
-    
-     // Let's reuse selectLookup logic if possible or just assert it exists
     if (lookupToSelect != null) {
-        // Find if the lookup contains our expected element
-        var lookups = myFixture.getLookupElementStrings();
-        assertNotNull("Lookup list should not be null", lookups);
+      var lookups = myFixture.getLookupElementStrings();
+      if (lookups != null) {
         assertTrue("Likely missing completion item: " + lookupToSelect + ". Found: " + lookups, lookups.contains(lookupToSelect));
+      } else {
+        // When a unique completion suggestion matches the prefix, IntelliJ auto-inserts it without showing a lookup list.
+        String editorText = myFixture.getEditor().getDocument().getText();
+        assertTrue("Lookup list was null and expected item '" + lookupToSelect + "' was not auto-inserted in editor:\n" + editorText,
+                   editorText.contains(lookupToSelect));
+      }
     }
   }
 }
--- a/third_party/src/test/java/com/jetbrains/dart/analysisServer/DartDotShorthandCompletionTest.java
+++ b/third_party/src/test/java/com/jetbrains/dart/analysisServer/DartDotShorthandCompletionTest.java
@@ -88,6 +88,18 @@ public class DartDotShorthandCompletionTest extends DartServerCompletionTest {
              }""");
   }
 
+  public void testDotNewWithWhitespaceAndCommentsArgCompletion() {
+    doTest("name: ",
+           """
+             class A {
+               A({String? name});
+             }
+             void f(A a) {}
+             main() {
+               f(.new /* comment */ (<caret>));
+             }""");
+  }
+
   private void doTest(String lookupToSelect, String text) {
     myFixture.configureByText("foo.dart", text);
     myFixture.doHighlighting();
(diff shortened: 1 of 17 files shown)


jakemac53:
Ok, I finally managed to squeeze more of an ELI5 description of this change out of gemini:


> ### Why our change in DartReferenceImpl is the sound architectural fix                                                                                                                                      
>                                                                                                                                                                                                               
>   In IntelliJ Platform's language API, when an AST node implements PsiReference, it is required to provide a method called getRangeInElement(). The strict design contract of getRangeInElement() is to       
>   separate the reference designator (the keyword or name being invoked) from the surrounding invocation syntax (the arguments and parentheses).                                                               
>                                                                                                                                                                                                               
>   Before our fix, DartReferenceImpl.java relied on an old assumption: “The callable reference designator will always be a child AST node of type DartReference (like Foo in new Foo()).”                      
>   When new Foo(x, y) was parsed, getRangeInElement() found the child node Foo and returned only Foo's range, cleanly excluding the arguments (x, y).                                                          
>                                                                                                                                                                                                               
>   When dot shorthand was added in 2025, that assumption was broken:                                                                                                                                           
>   In .new(x, y), .new consists only of primitive lexer leaf tokens (the dot . and keyword 'new') without a wrapping AST reference node. When getRangeInElement() found zero child reference nodes, it fell    
>   back to returning the text range of the entire expression—unwittingly including the argument list (x, y).                                                                                                   
>                                                                                                                                                                                                               
>   ### Summary                                                                                                                                                                                                 
>                                                                                                                                                                                                               
>   By changing DartReferenceImpl.java to explicitly terminate the reference range right before DartArguments, we are not applying a band-aid—we are fixing a flaw in the implementation of IntelliJ's          
>   PsiReference contract. This ensures that across all current and future constructor or method syntax variants, argument lists will never be misclassified as part of a symbol's identifier range.

third_party/src/test/java/com/jetbrains/dart/analysisServer/DartDotShorthandCompletionTest.java
```diff
@@ -40,22 +40,80 @@ void f(E e) {}
...
+             class A {
+               A({String? name});
+             }
+             void f(A a) {}
+             main() {
+               f(.new(<caret>));
+             }""");
+  }
+
+  public void testDotNamedPartialArgCompletion() {
```
pq, line 67: I'll have to look closer at the grammar but I wonder if it wouldn't be good to add a test with primary named constructor too?

pq:
Awesome. This looks a lot better to me; I'd still love a look over from Brian since he's got a good eye for grammar nuances. Also: maybe some test coverage w/ primary constructors?

Assuming he doesn't catch any snags, LGTM!

bwilkerson (commented):
Based on `Dart.bnf`, the grammar being accepted wasn't changed, just some details about the way the parser generator works. Not being familiar with their parser generator, I don't know how the changes impacted the generated code. I don't think I have anything to contribute to this review. Sorry.

pq:
NP. Thanks for taking a look, Brian!

pq (approved):
In conversation you mentioned this may fix other untested aspects of our dot-shorthand support. It'd be cool while we're here to see if we can validate this and add some new tests? (Maybe worth a quick query anyway.)

Needless to say: not blocking.

Thanks for this!

third_party/src/test/java/com/jetbrains/dart/analysisServer/DartDotShorthandCompletionTest.java
jakemac53: done

jakemac53:
> In conversation you mentioned this may fix other untested aspects of our dot-shorthand support. It'd be cool while we're here to see if we can validate this and add some new tests? (Maybe worth a quick query anyway.)
> 
> Needless to say: not blocking.
> 
> Thanks for this!

It wasn't able to find anything else that was actually broken 🤷

Issue #462, opened 2026-06-11
3 crashes, possibly related to the new constructor syntax


johnniwinther:
**Describe the bug**
Three crashes occurred. I don't know what triggered it but the code is using the new constructor syntax from the primary-constructors feature, so it might be related to that.

**Version information**
 - Dart plugin version: 507.0.0-dev.20260608-7ea7a1c
 - IDE platform version: IntelliJ IDEA 2026.1.3, build #IU-261.25134.95
 - OS: Windows 11

**Error messages / Logs / Other context**
1)
```
java.lang.Throwable: [Split debugger] RunContentDescriptor should not be used in split mode from XDebugSession. XDebugSession.getRunContentDescriptor is deprecated, see the javadoc for details
	at com.intellij.openapi.diagnostic.Logger.error(Logger.java:375)
	at com.intellij.xdebugger.impl.XDebugSessionImpl.getRunContentDescriptor(XDebugSessionImpl.kt:267)
	at com.jetbrains.lang.dart.ide.runner.DartRunner.doExecuteDartDebug(DartRunner.kt:140)
	at com.jetbrains.lang.dart.ide.runner.DartRunner.doExecute(DartRunner.kt:68)
	at com.intellij.execution.runners.GenericProgramRunner.execute$lambda$0(GenericProgramRunner.kt:20)
	at com.intellij.execution.impl.ExecutionManagerImpl.startRunProfile$lambda$0(ExecutionManagerImpl.kt:275)
	at com.intellij.execution.impl.ExecutionManagerImpl.doStartRunProfile$lambda$1(ExecutionManagerImpl.kt:342)
	at com.intellij.openapi.application.TransactionGuardImpl.runWithWritingAllowed(TransactionGuardImpl.java:236)
	at com.intellij.openapi.application.TransactionGuardImpl.access$100(TransactionGuardImpl.java:22)
	at com.intellij.openapi.application.TransactionGuardImpl$1.run(TransactionGuardImpl.java:198)
	at com.intellij.openapi.application.impl.AppImplKt$runnableUnitFunction$1.invoke(appImpl.kt:124)
	at com.intellij.openapi.application.impl.AppImplKt$runnableUnitFunction$1.invoke(appImpl.kt:124)
	at com.intellij.platform.locking.impl.NestedLocksThreadingSupport.runWriteIntentReadAction(NestedLocksThreadingSupport.kt:705)
[...]

PR #578, opened 2026-08-05
Fix read access exception in DartAnalysisServerService.updateVisibleFiles

jakemac53:
Part of https://github.com/flutter/dart-intellij-third-party/issues/462 - addresses the 2nd crash.

Wrap updateVisibleFiles in runReadAction instead of assertReadAccessAllowed so that event listeners invoked on the event dispatch thread outside of an active read action do not throw thread access exceptions.

### DIFF
--- a/third_party/src/main/java/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
+++ b/third_party/src/main/java/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
@@ -967,24 +967,24 @@ public final class DartAnalysisServerService implements Disposable {
   }
 
   public void updateVisibleFiles() {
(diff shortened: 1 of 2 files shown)


third_party/src/main/java/com/jetbrains/lang/dart/analyzer/DartAnalysisServerService.java
```diff
@@ -967,24 +967,24 @@ public boolean isInIncludedRoots(final @Nullable VirtualFile vFile) {
   }
 
   public void updateVisibleFiles() {
-    ApplicationManager.getApplication().assertReadAccessAllowed();
```
pq, line 970: I have to wonder why this was ever exception-throwing but the move feels good to me. Thanks!

pq (approved):

Issue #597, opened 2026-08-11
App start black ~30s and stops ~30s after upgrade to 508.0


robertosw:
**Describe the bug**
Launching from IntelliJ IDEA IDE using "Flutter" run configuration (debug and profile): Window starts with OS title bar but content is black for ~30s, then works. Stopping (X on OS window) closes window visually, but IDE takes ~30s to register this and allow starting another window. Release mode starts fast, but stops slow. All modes are fast (as expected) from the terminal.

- Plugin versions: bad Dart: 508, bad Flutter 95.0; good Dart 507.0, good Flutter 94.0.
     - New Flutter + new Dart: slow
     - Old Flutter + new Dart: slow
     - Old Flutter + old Dart: fast                                                                 
     - New Flutter + old Dart: not tested
 - Reproducible in vanilla Flutter app with no packages on Linux desktop. 
 - Workaround: downgrade both plugins one version.

**Environment**
IntelliJ IDEA 2026.2.0 & 2026.2.1
Linux desktop (Debian 12, kernel 6.12), Intel Xe iGPU

```shell
$ flutter doctor -v
[✓] Flutter (Channel stable, 3.38.5, on openSUSE Leap 16.0 6.12.0-160000.36-default, locale de_DE.UTF-8) [32ms]
    • Flutter version 3.38.5 on channel stable at /home/robert/Dokumente/flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision f6ff1529fd (vor 8 Monaten), 2025-12-11 11:50:07 -0500
    • Engine revision 1527ae0ec5
    • Dart version 3.10.4
    • DevTools version 2.51.1
    • Feature flags: enable-web, enable-linux-desktop, enable-macos-desktop, enable-windows-desktop, enable-android, enable-ios, cli-animations, enable-native-assets, omit-legacy-version-file, enable-lldb-debugging

[✓] Android toolchain - develop for Android devices (Android SDK version 36.1.0) [2,7s]
    • Android SDK at /home/robert/Android/Sdk
    • Emulator version 36.6.11.0 (build_id 15507667) (CL:N/A)
    • Platform android-36, build-tools 36.1.0
    • Java binary at: /home/robert/.local/share/JetBrains/Toolbox/apps/android-studio/jbr/bin/java
      This is the JDK bundled with the latest Android Studio installation on this machine.
      To manually set the JDK path, use: `flutter config --jdk-dir="path/to/jdk"`.
[...]

PR #599, opened 2026-08-12
Add sdk guard for --ping-interval.

JavadAsadi:
## Summary

Relates to https://github.com/flutter/dart-intellij-third-party/issues/597

since --ping interval option was added on sdk 3.11, we need to make sure we don't add this to the command with older sdk versions.

### DIFF
--- a/third_party/src/main/java/com/jetbrains/lang/dart/ide/toolingDaemon/DartToolingDaemonCommandLine.kt
+++ b/third_party/src/main/java/com/jetbrains/lang/dart/ide/toolingDaemon/DartToolingDaemonCommandLine.kt
@@ -36,8 +36,12 @@ internal fun createDtdCommandLine(sdk: DartSdk, pingInterval: Int = 15): General
   return commandLine
 }
 
-internal fun supportsDtdPingInterval(sdkVersion: String): Boolean =
-  sdkVersion.firstOrNull()?.isDigit() == true &&
-  DartSdkUpdateChecker.compareDartSdkVersions(sdkVersion, MIN_DTD_PING_INTERVAL_SDK_VERSION) >= 0
+internal fun supportsDtdPingInterval(sdkVersion: String): Boolean {
+  if (sdkVersion.isBlank() || !sdkVersion.first().isDigit()) {
+    return false
+  }
+
+  return DartSdkUpdateChecker.compareDartSdkVersions(sdkVersion, MIN_DTD_PING_INTERVAL_SDK_VERSION) >= 0
+}
 
 internal fun isDtdCommandLine(text: String): Boolean = DTD_COMMAND_LINE_PATTERN.matches(text)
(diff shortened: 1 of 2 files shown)


third_party/src/main/java/com/jetbrains/lang/dart/ide/toolingDaemon/DartToolingDaemonCommandLine.kt
```diff
@@ -18,18 +19,25 @@ private val DTD_COMMAND_LINE_PARAMETERS = listOf(
...
-  commandLine.addParameter("$DTD_PING_INTERVAL_PARAMETER_PREFIX$pingInterval")
+  if (supportsDtdPingInterval(sdk.version)) {
+    commandLine.addParameter("$DTD_PING_INTERVAL_PARAMETER_PREFIX$pingInterval")
+  }
   Analytics.updateEnvironment(commandLine)
   return commandLine
 }
 
+internal fun supportsDtdPingInterval(sdkVersion: String): Boolean =
+  sdkVersion.firstOrNull()?.isDigit() == true &&
```
helin24: is this meant to be a null check? Seems like an odd condition.

helin24 (approved):
LGTM - sorry i didn't think about the version check when we first added this.

third_party/src/main/java/com/jetbrains/lang/dart/ide/toolingDaemon/DartToolingDaemonCommandLine.kt
pq: It looks odd but I think it's right. Really you only want this to return true if the SDK version is non-null and is a digit. This does that; though it is admittedly a bit hard to read.

pq (approved):

third_party/src/main/java/com/jetbrains/lang/dart/ide/toolingDaemon/DartToolingDaemonCommandLine.kt
JavadAsadi: I can fix that if it's hard to read.

To clarify the non-digit cases, it might be "unkown".

```
public final class DartSdk {
  public static final String DART_SDK_LIB_NAME = "Dart SDK";
  private static final String UNKNOWN_VERSION = "unknown";
```

JavadAsadi: changed with this:

```
internal fun supportsDtdPingInterval(sdkVersion: String): Boolean {
  if (sdkVersion.isBlank() || !sdkVersion.first().isDigit()) {
    return false
  }

  return DartSdkUpdateChecker.compareDartSdkVersions(sdkVersion, MIN_DTD_PING_INTERVAL_SDK_VERSION) >= 0
}```

pq: LGTM. Thanks!

pq:
I haven't done any digging but I expect the failing integration test is unrelated. (Almost certain.)

Re-running now but @helin24, feel free to merge either way

helin24:
I'll investigate the test further.

PR #642, opened 2026-09-03
Add UserTag Support in VM Service Drivers

JavadAsadi:
## Summary

  Upgrades the VM Service drivers from protocol 4.3 to 4.4 by:

  - Adding UserTag to InstanceKind.
  - Exposing the UserTag label through Instance and InstanceRef.
  - Updating the supported protocol version to 4.4.

  ## Testing

  - Added a unit test verifying that both Instance and InstanceRef decode the UserTag kind and label from VM Service JSON.
  - Existing VmServiceTest passes.
  - Manually debugged a Dart UserTag in the sandbox IDE and confirmed the VM payload contains kind: UserTag and label: manual-test-tag.

### DIFF
--- a/third_party/src/main/java/com/jetbrains/lang/dart/ide/runner/server/vmService/vmServiceDrivers/service/VmService.java
+++ b/third_party/src/main/java/com/jetbrains/lang/dart/ide/runner/server/vmService/vmServiceDrivers/service/VmService.java
@@ -82,7 +82,7 @@ public class VmService extends VmServiceBase {
   /**
    * The minor version number of the protocol supported by this client.
    */
-  public static final int versionMinor = 3;
+  public static final int versionMinor = 4;
 
   /**
    * The [addBreakpoint] RPC is used to add a breakpoint at a specific line of some script.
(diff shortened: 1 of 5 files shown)


helin24:
(Sorry if this was already discussed in chat) Can you add more context on why we need UserTag?

pq:
> Can you add more context on why we need UserTag?

Is the goal of this change just to keep up w/ the newest protocol version (4.4)?

I'm curious, in general, if we should consider some automation around (stale) version detection?

JavadAsadi:
> (Sorry if this was already discussed in chat) Can you add more context on why we need UserTag?



> > Can you add more context on why we need UserTag?
> 
> Is the goal of this change just to keep up w/ the newest protocol version (4.4)?



My intention for this pr is to have the first step towards updating the Vmservice to match with

https://github.com/dart-lang/sdk/blob/main/runtime/vm/service/service.md

Right now the latest version is 4.22, so we are a few versions behind.

UserTag is a kind of [Instance](https://github.com/dart-lang/sdk/blob/main/runtime/vm/service/service.md#instance).
I don't know how this information helps the developer while using the VM service.

I manually tested it by puting break points in the code and saw that the kind was "kind: "UserTag".

> 
> I'm curious, in general, if we should consider some automation around (stale) version detection?

I'm working on the skill to match the version to the latest version, this is the first output of that skill. 
I think once the skill is ready then it's a good idea to have the automation.

pq (approved):
LGTM!

helin24 (approved):
Thanks!

PR #643, opened 2026-09-04
Remove disambiguation for find usages

helin24:
This is a follow up to https://github.com/flutter/dart-intellij-third-party/pull/623. The LSP change introduced a problem in find usages where a pop up menu was shown on find usages invocation:
image

The first line is from LSP, which just indicates the location where the cursor is. The second line is from an IJ platform class, which finds a PSI target unless we override this behavior.

This fix overrides the IJ platform class `TargetElementEvaluatorEx2` to not return any targets if LSP is enabled. Then only one target is found when find usages is triggered, and there is no menu in between.

To test this manually:
- Make sure that when Find usages is triggered, the usages show up in tool window immediately instead of requiring the user to select a target first:
image
- Also ensure that Find usages behavior remains the same when you uncheck experimental LSP and use the legacy implementation instead.

This problem wasn't reproducing everywhere for me, but I was able to consistently see this happening in the Dart SDK file `sdk/pkg/analyzer/lib/src/generated/resolver.dart`.

---

<details>

  `dart format`.

</details>

### DIFF
--- a/third_party/CHANGELOG.md
+++ b/third_party/CHANGELOG.md
@@ -7,6 +7,7 @@
 ### Removed
 
 ### Fixed
+- Remove ambiguous "Find Usages Of" popup when invoking Find Usages with LSP references enabled (#643)
 
 ## 509.0.0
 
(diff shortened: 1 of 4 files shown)


pq:
This is just a temporary fix right and will go away when we're fully onto LSP? (If so, maybe a comment to that effect -- or better still, do you have a punch list where you're tracking things like this?)

pq (approved):

helin24:
Yes, this is a temporary fix, but part of it won't be removable until we are getting PSI elements through LSP. This means the latter part will require we move functionality outside of find usages to LSP as well. The closest thing to a punch list is probably tracking which features are in experimental mode, since we will remove plugin code after moving features out of experimental mode.

PR #669, opened 2026-09-17
Add getAsLong in Element.java

JavadAsadi:
## Summary

This is a simple refactoring PR.

Element.java already had getAsString, getAsInt, and ... . But missed getAsLong. This caused every other class to directly handle the null check.

So now null checking for getAsLong is being handled in Element.java just like getAsInt, getAsString, getAsBoolean.

This pr is also is preparing for adding PerfettoCpuSamples (VmService 4.6).


pq (approved):
Nice! 🧹

helin24 (approved):

PR #663, opened 2026-09-15
Migrate Closing Labels to LSP over Legacy bridge (#400)

ranbeuer:
Migrates Dart Closing Labels (`dart/textDocument/publishClosingLabels`) to LSP over the LSP-over-Legacy (LoL) bridge, resolving [#400](https://github.com/flutter/dart-intellij-third-party/issues/400).
    
Previously, Closing Labels and Outline were grouped together. Following upstream Dart SDK changes ([dart-lang/sdk#64200](https://github.com/dart-lang/sdk/issues/64200): CL 548160 and CL 547223, first available in Dart SDK `3.14.0-219.0.dev`), the legacy Dart Analysis Server (DAS) now suppresses legacy `analysis.closingLabels` notifications and sends reverse `dart/textDocument/publishClosingLabels` notifications over the LSP bridge when `experimental.closingLabels` is advertised in `server.setClientCapabilities`.
    
---
    
## Changes
    
- **`DartAnalysisServerService.java`**:
  - Added `MIN_LSP_CLOSING_LABELS_SDK_VERSION = "3.14.0-219.0.dev"`.
  - Added `isDartSdkVersionSufficientForLspClosingLabels(...)` and `isLspClosingLabelsEnabled(...)`.
  - Updated `buildLspCapabilities(...)` to advertise `"experimental": { "closingLabels": {} }` when LSP Closing Labels is enabled.
  - Added `onLspClosingLabelsUpdated(String filePathUri, List<ClosingLabel> labels)` to update `DartServerData.computedClosingLabels(...)` and trigger declarative inlay hint refreshes.
  - Left legacy `AnalysisService.CLOSING_LABELS` subscription in `updateSubscriptions()` untouched so older SDKs (`< 3.14.0-219.0.dev`) or sessions with experimental LSP features disabled fall back seamlessly to legacy DAS notifications.
- **`DartLspClosingLabelsConverter.kt`** (new):
  - Converts 0-indexed LSP `Range`s from `dart/textDocument/publishClosingLabels` into IntelliJ document `offset` and `length` (`ClosingLabel`) inside a read action.
- **`DartBridgeLspServer.kt`**:
  - Intercepts `dart/textDocument/publishClosingLabels` reverse notifications in `forwardNotificationToClient(...)`, converts the labels via `DartLspClosingLabelsConverter`, and forwards them to `DartAnalysisServerService.onLspClosingLabelsUpdated(...)`.
- **Unit Tests**:
[...]

PR #670, opened 2026-09-17
[verifier] improve breakage visibility in the CI

### LINKED ISSUE #668: [verifier] make new errors easier to see in the CI
For example: https://github.com/flutter/dart-intellij-third-party/actions/runs/35036138172/job/104606603022?pr=663

What's the new failure?

(FWIW: I think this is a regression.)

pq:
**The rub.** When `verifyPlugin` fails, the CI log is a wall of pre-existing noise and the one line that matters is buried. Worse, the baseline check that would have pointed at the new issue never ran, because:

- `Check baselines` had no `if:` guard, so it was skipped exactly when the verifier failed.
- All four gradle tasks shared one `run:` block, so `set -e` aborted the rest after the first failure.
- The baseline loop hardcoded `253 261 262`, but `ides { recommended() }` resolves **four** IDEs.
  Branch 263 had no baseline and was never checked.
- A blanket `grep -v com.intellij.platform.dartlsp` discarded real compatibility problems
  in vendored code.

Net effect on [PR #663](https://github.com/flutter/dart-intellij-third-party/pull/663): five `NoSuchMethodError` risks against `IU-263.4732.28` (lsp4j 1.0.0 changed `Diagnostic.message` and `DocumentFilter.pattern` to `Either<...>`) and nothing in the log said so.

## New

The new output makes these new errors more obvious.

image

## Old

Compare that with:

image

## Change

- Split the gradle tasks into separate steps; everything after `verifyPlugin` runs under
  `if: ${{ !cancelled() }}` so the diagnosis prints even when the verifier fails.
- New shared script `third_party/tool/check_verifier_baselines.sh` with `check` / `update`
  modes. `update_baselines.sh` and `.bat` now delegate to it, so the logic can't fork.
  It iterates the baseline directories that exist instead of a hardcoded list, and adds a
  baseline for 263.
- New issues are emitted as `::error` annotations and a job-summary table, so they surface
  at the top of the run rather than in the log.
- Hard-fails on an unparseable report, an unknown report section, or a missing baseline —
  previously each of these produced a silent green build.

Severity policy is unchanged: `failureLevel` in `build.gradle.kts` still decides what is *severe*; the script only decides what is *new*. dartlsp filtering behavior is unchanged, apart from no longer swallowing compatibility problems.

## Why the baseline format changed

Baselines go from bare bullet text:

```
* Deprecated method ...DaemonCodeAnalyzer.restart() ... is invoked in ...
```

to `<section><TAB><issue>`:

```
Deprecated API usages	Deprecated method ...DaemonCodeAnalyzer.restart() ... is invoked in ...
```

The old format threw away the report section each entry came from, which made it impossible to scope the vendored code filter. Ignoring `dartlsp` entries is only ever correct inside the informational sections (Deprecated / Experimental / Internal / Override-only); applying it everywhere is what hid the five lsp4j compatibility problems. Carrying the section makes the filter precise, lets the diff be grouped by section in the annotations and summary, and makes baseline files readable on their own.

This is a pure re-record — the entry counts for 253/261/262 are unchanged (17/25/47).

## Not in scope

The lsp4j breakages themselves. This PR is about making the failure legible, not green; branch 263 will still report those five problems until they're fixed (tracked separately). (Will track with a new issue.)

Fixes: #668 

---

<details>

  `dart format`.

</details>


helin24 (approved):
Thanks!

Issue #622, opened 2026-08-21
Inlay hints stage 2: per-category checkboxes (needs configuration transport over LSP-over-Legacy)


ralph-bergmann:
Follow-up to #617 (see the two-stage proposal discussed there): stage 1 ships two kind-level checkboxes — *Settings | Editor | Inlay Hints → Parameter names → Dart* and *… → Types → Dart* — that filter client-side by `InlayHint.kind`.

Stage 2 adds one checkbox per server-side hint category, as nested `<option>` checkboxes under those same two providers:

- under *Types*: variable types, return types, parameter types, type arguments, dot shorthand types
- under *Parameter names*: an "only for literal arguments" option (the server's `parameterNames` setting is three-valued: `none | literal | all`, default `all`)

This granularity cannot be filtered on the client: on the wire a hint only carries `kind: Type|Parameter`, so the server has to stop producing the disabled categories. The server already honors the `dart.inlayHints` configuration in the shared `InlayHintHandler` — what's missing is only the transport of that configuration over LSP-over-Legacy.

The approach I propose is the standard LSP configuration flow wrapped in the existing `lsp.handle` envelope: the client sends `workspace/didChangeConfiguration`, and the server pulls the `dart` section via a `workspace/configuration` request through its existing reverse-`lsp.handle` channel (enabled by the `lspCapabilities` from #614). That way a later full-LSP migration only removes the envelope — the plugin's config building, the server behavior and the tests of the behavior all carry over. Two small SDK changes are needed first (the `lsp.handle` handler currently accepts only LSP *requests*, not notifications, and the configuration fetch is registered LSP-only); I'll file the SDK issue separately and link it here.

Part of the #207 migration; original feature request: #159.

ralph-bergmann:
SDK-side issue filed: dart-lang/sdk#64101 (accept LSP notifications in the `lsp.handle` envelope + share the configuration fetch with the legacy server).

PR #679, opened 2026-09-19
Per-category inlay hint options, configured over LSP-over-Legacy

### LINKED ISSUE #64101: Support client configuration (workspace/didChangeConfiguration + workspace/configuration) over LSP-over-Legacy
With #64061 landed (7c18d1fa0e5), the Dart IntelliJ plugin shows inlay hints over LSP-over-Legacy (flutter/dart-intellij-third-party#617). The follow-up there is per-category settings — variable types, return types, parameter types, type arguments, dot shorthand types, parameter-name mode (flutter/dart-intellij-third-party#622). That granularity is server-side by design: the shared `InlayHintHandler` already honors `server.lspClientConfiguration.global.inlayHints` (from #60326), but a legacy client currently has no way to get configuration into `lspClientConfiguration`.

Most of the plumbing already exists on the legacy server:

- it stores LSP client capabilities from `server.setClientCapabilities`'s `lspCapabilities` field (`legacy_analysis_server.dart`, `_editorClientCapabilities`), and
- it implements `sendLspRequest` (server→client requests wrapped in `lsp.handle`), so the `workspace/configuration` pull that `fetchClientConfigurationAndPerformDynamicRegistration` does today would work over legacy as-is, gated by `editorClientCapabilities?.configuration` exactly like on the LSP server.

Two gaps remain:

1. `LspOverLegacyHandler` only parses `RequestMessage`s — an LSP *notification* such as `workspace/didChangeConfiguration` (no `id`) is answered with `INVALID_PARAMETER`, so a client can't even tell the server that settings changed.
2. `WorkspaceDidChangeConfigurationMessageHandler` is registered in `lspHandlerGenerators` (LSP-only), and `fetchClientConfigurationAndPerformDynamicRegistration` lives on `LspAnalysisServer`, mixing the shareable part (fetch the global `dart` section, `lspClientConfiguration.replace(...)`) with LSP-only concerns (per-workspace-folder config items, `capabilitiesComputer.performDynamicRegistration()`).

Proposal: accept `NotificationMessage`s in the `lsp.handle` envelope (acknowledging with an empty `LspHandleResult`), move the configuration fetch down to the base `AnalysisServer` (the legacy server would request only the global `dart` section — the settings relevant here, like `dart.inlayHints`, are workspace-level only anyway), and register the `didChangeConfiguration` handler as shared. Using the standard LSP configuration flow rather than a new legacy request keeps the code that clients and server need identical to full LSP, so nothing has to be revisited when the IntelliJ plugin finishes its migration.

As with #64061, happy to do the CL if this direction sounds right.

ralph-bergmann:
Stage 2 of #622 (follow-up to #617): one checkbox per server-side inlay hint category, nested under the two Stage-1 providers in *Settings | Editor | Inlay Hints*:

- **Types → Dart:** Variable types, Return types, Parameter types, Type arguments, Dot shorthand types (all on by default, mirroring the server defaults)
- **Parameter names → Dart:** Only for literal arguments (off by default = the server's `all`)

These categories cannot be filtered on the client (a hint only carries `kind: Type|Parameter`), so the plugin now sends the user's settings to the analysis server as the LSP `dart` configuration section — through the existing legacy transport, using the standard LSP configuration flow.

> **Depends on dart-lang/sdk CL [546841](https://dart-review.googlesource.com/c/sdk/+/546841)** (fixes dart-lang/sdk#64101, in review, protocol 1.41.0). Until it lands, the new options are visible but have no effect with any released SDK. Before merging: re-check that the landed CL still reports protocol `1.41.0`, replace "(requires a recent Dart SDK)" in the six option descriptions and the CHANGELOG line with the first dev SDK that contains it, and confirm the visible-but-inert UX below the floor (the static EP XML cannot hide options).

## How it works

1. **Capability.** `server.setClientCapabilities` now always carries `lspCapabilities.workspace.configuration = true`. Capabilities describe the client, and the server protocol version is not yet known when they are sent (`server.connected` arrives later); a server without the CL parses and ignores this standard LSP field.
2. **Pull.** A server with the CL answers `setClientCapabilities`, then sends `workspace/configuration` (items: the global `dart` section) as a reverse `lsp.handle` request, and holds later LSP-over-Legacy messages until it is answered. The plugin answers with `DartLspInlayHintsConfiguration.buildDartSection()` — `{"inlayHints": {variableTypes/returnTypes/parameterTypes/typeArguments/dotShorthandTypes: {enabled: bool}, parameterNames: {enabled: "none"|"literal"|"all"}}}`; parent checkbox off ⇒ all its categories off. The reply is guaranteed exactly once, also if the supplier throws (a missing reply would block the server's initialization).
3. **Push.** `DartLspInlayHintSupport.shouldAskServerForInlayHints` (already called before every inlay-hint request cycle) compares the current section with the one the server last read; on drift it sends `workspace/didChangeConfiguration` once, as a legacy `lsp.notification` (`{"event": "lsp.notification", "params": {"lspNotification": {...}}}`, no id). The server then pulls again (step 2). Sending is gated on the legacy **protocol** version from `server.connected` ≥ `1.41.0` — a protocol rather than an SDK version because the CL has not landed and no dev tag exists yet; the answer path (step 2) is deliberately not gated: every server that asks gets the real settings. State lives in `DartLspConfigurationSync` (per project, lock-guarded; reset on server restart).
4. **Refresh.** The server does not send `workspace/inlayHint/refresh`, and the vendored `LspHighlightingCache` re-requests only when `PsiModificationTracker.modificationCount` changed — `LspServerManagerImpl.refreshInlayHints` alone (a daemon restart) shows the cached hints again. So after the server has read the new section the plugin calls `PsiManager.dropPsiCaches()` + `LspServerManagerImpl.refreshInlayHints(project)` on the EDT. **This needs your sign-off:** it is a project-wide PSI-cache drop per applied settings change (a rare user action; in the sandbox it was a brief re-highlight). The cache's `clearCache()` is `internal` in `thirdPartySrc/platform-lsp`, which I did not touch; the clean fix would be a small invalidation hook there — happy to do that instead if you prefer changing the vendored JetBrains code.

Transport decision: the LSP configuration flow inside the legacy envelope (instead of a new legacy request or `analysis.updateOptions`) keeps the code that has to be revisited for the full LSP migration minimal — the config building, the server behaviour and the tests carry over, only the envelope goes away — and since the pull replaces the server's `lspClientConfiguration` wholesale, every future `dart.*` option rides the same flow (only `buildDartSection()` grows).

## Vendored client changes (`thirdPartySrc/analysisServer/**`)

Following the precedent of #615 (`RequestUtilities.java`) and the existing `workspace/applyEdit` dispatch in `RemoteAnalysisServerImpl.java`:
- `RemoteAnalysisServerImpl`: dispatch `workspace/configuration` → new abstract `AnalysisServer.lsp_workspaceConfiguration(sections, consumer)` (implemented in `DartAnalysisServerImpl.kt`, so the vendored client stays free of IntelliJ settings knowledge); answered-exactly-once guard; `sendNotificationToServer` (no consumer registration); the apply-edit reply now echoes the LSP id verbatim (it was stringified; the server matches on the legacy id, so no behaviour change).
- `RequestUtilities.generateLSPResponse`: shared `lspResponse` envelope for apply-edit and configuration.
- New `DartLspWorkspaceConfigurationConsumer`.

`thirdPartySrc/platform-lsp/**` is untouched. (`src/main` already imports `com.intellij.platform.dartlsp.impl` in the hierarchy providers; `LspServerManagerImpl.refreshInlayHints` is reachable because both trees share one source set.)

## Threading

The push runs on the inlay-hint background thread inside a read action; the wire write is an unbounded queue put, not blocking I/O. The reply runs on the DAS response-reader thread (`buildDartSection` reads only `DeclarativeInlayHintsSettings` and the provider registry — no PSI, no EDT). Only the refresh touches the EDT, via `invokeLater(ModalityState.nonModal())`. No lock is held across any of it.

## Testing

- 83 unit tests green (`lsp.*`, `analyzer.*`, `hints.*`): registration of the options, section building, the reverse-request wire shape (captured JSON through the real `RemoteAnalysisServerImpl`), the push state machine (single toggle, back-and-forth, server-initiated pull, restart, send failure, old server never notified), and the capability.
- Manual, with an SDK built from CL 546841 ps5 and protocol logs (details on the CL): (a) the current Marketplace plugin + new SDK → server behaves exactly as before; (b) this branch + new SDK → startup pull, one notification + one pull per change, per-category and parent toggles take effect without editing the file, restart re-syncs; (c) this branch + Dart 3.14.0-217.0.dev (protocol 1.40.1) → capability ignored, no notification, options inert, no errors.
- Plugin verifier: no new issues from this change (only the known lambda-renumbering and 262/263 baseline artefacts).

## Known limitations / follow-ups

- Below the version floor the six options are visible but do nothing (see the note at the top).
- An `inlayHint` request issued in the same instant as the notification is answered with the previous configuration; the request after the pull wins (documented in the CL).
- The refresh currently issues two identical `inlayHint` requests per open file a few milliseconds apart (`dropPsiCaches` and the daemon restart each trigger a pass).
- Pre-existing, made more relevant by the server now waiting on a reverse answer: `processLspRequestFromServer` dereferences `lspMessage.id`/`method` unguarded and silently drops unknown reverse methods — follow-up: #680.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ralph-bergmann:
Inlay Hint settings now:

SCR-20260919-qxqc

third_party/thirdPartySrc/analysisServer/com/google/dart/server/internal/remote/RemoteAnalysisServerImpl.java
```diff
@@ -964,23 +969,84 @@ public void workspaceEditApplied(DartLspApplyWorkspaceEditResult result) {
...
-        sendResponseToServer(responseElement);
+        sendResponseToServer(RequestUtilities.generateLSPResponse(dasRequestId, lspRequestId, lspResultElement));
       }
     };
 
-    lsp_workspaceApplyEdit(workspaceEditParams, consumer);
+    try {
+      lsp_workspaceConfiguration(sections, consumer);
+    }
+    catch (RuntimeException e) {
```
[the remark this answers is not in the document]
ralph-bergmann: Done in 4ef891c10: the fallback now catches `Throwable`. You are right that the reader loop swallows Errors as well (a `NoClassDefFoundError` during a plugin unload would have left the request unanswered), so the answer-first fallback now covers them too, with a test for the Error path.

helin24 (approved):
This all looks good to me. Dropping caches on change is fine; I'm expecting those options won't get changed much.

I noticed that feature gating is on analysis server version instead of SDK version; it may be nicer to do SDK version just to be similar to other LSP changes and not be a point of confusion when we change to full LSP; am I correct that this versioning won't be needed when we move to full LSP?

ralph-bergmann:
@helin24 Switched the gate to the SDK version in 471f15f21: `MIN_LSP_INLAY_HINTS_CONFIGURATION_SDK_VERSION = "3.14.0-258.0.dev"`, the first dev SDK that contains dart-lang/sdk@6700ccc4316, checked with `DartSdkUpdateChecker.compareDartSdkVersions` like the other LSP gates. The protocol version was only a stand-in while no dev SDK contained the change.

And yes, with full LSP this versioning is not needed:
- There is no legacy protocol version anymore, and the LSP server has handled `workspace/configuration` and `workspace/didChangeConfiguration` since Dart 2.9, and the per-category `dart.inlayHints` settings since Dart 3.10 (dot shorthand types since 3.11). That is all well below the 3.14.0-139.0.dev that LSP inlay hints already require.
- What carries over is `DartLspInlayHintsConfiguration.buildDartSection()`, returned from `DartLspServerDescriptor.getWorkspaceConfiguration(item)` for the `dart` section (the platform LSP client already answers `workspace/configuration` through that hook), plus the drift check, which would send `workspace/didChangeConfiguration` through `LspServer.sendNotification`.
- What goes away is the legacy plumbing: the `lsp.notification` sender, the reverse-request dispatch in `RemoteAnalysisServerImpl`, and this gate.

helin24:
Thanks @ralph-bergmann  !

@pq this is ready for your review.

third_party/thirdPartySrc/analysisServer/com/google/dart/server/generated/AnalysisServer.java
```diff
@@ -10,6 +10,7 @@
 
 import com.google.dart.server.*;
 import org.dartlang.analysis.server.protocol.*;
+import org.jetbrains.annotations.Nullable;
```
pq: If this file is still generated, I don't think we want to make modifications here?

pq (changes requested):

third_party/thirdPartySrc/analysisServer/com/google/dart/server/generated/AnalysisServer.java
helin24: Hmmm good point. I will investigate whether we should generate and update this file separately and land a change before we update here.

helin24: After a little more investigation, my inclination is to just leave this change.

This file has already drifted from what would be generated, and it's only for communication with the legacy DAS, which means we are aiming to remove this soon. I don't think we've ever had a solid process for syncing this up to the SDK code, and we probably shouldn't bother to now.

